PR #12, opened 2015-12-22
WEB: Translation updates for the German website

rootfather:
This request contains some minor fixes and additions to the German translation of the upcoming, new website. I'll add further commits to this request as translation progresses.


rootfather:
I think all issues with encoding are solved by now. I also found a typo left in the Sherlock announcement.

criezy:
I have manually cherry-picked the commits, dropping the first two commits relating to the Sherlock announcement (8628f1c and 2abbc53). I think this should be fine, but if you notice something missing let me know.

PR #13, opened 2016-01-17
WEB: Sync German translation to latest changes

rootfather:
This 2 commits update the German translation of the website to the current state of the English website (now sync with commit cbda272). It also localizes the news date format to our local format.


sev-:
Great! Applied, thanks

PR #16, opened 2016-01-21
WEB: Fix language switching

RichieSams:
As mentioned in f3c2203e2fc179e3f384e0499c34dcff63e792c1,
the language switching was always 1 behind. As in, the
language would be the last language you selected.

The first problem is that some of the smarty config variables
were being set after the call to _smarty->config_load(). Since
the smarty instance is cached, it was using the last set of variables.

I don't know why the production web site isn't throwing errors. My
personal set up wouldn't load until I could set _smarty->compile_dir
the first time. Perhaps the language update was done on the fly, so
smarty was still cached.

The next potential problem is persistence with $_SESSION.
The session is never initialized with session_start(). However,
the production php.ini may have session auto start enabled.

### DIFF
--- a/include/Controller.php
+++ b/include/Controller.php
@@ -29,7 +29,16 @@ class Controller {
 		$Smarty = $this->_smarty;
 
 		global $lang;
-
+		
+		/* Configure smarty. */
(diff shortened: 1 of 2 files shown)


sev-:
Thanks!

PR #20, opened 2016-02-01
WEB: Update Maemo downloads

tsoliman:
The 1.7.0 build for Maemo was uploaded 2015-09-14.
I didn't realize the download page still says 1.6.0 is the latest.

I am not sure if this is enough because I have no experience with scummvm-web. Please let me know if there is anything else that is required. I will update the wiki when this is resolved.


criezy:
Thank you (and yes this is enough - although I stil have to pull the changes on the website which I will do later tonight after a few more changes). I have manually merged this after amending to change the size for consistency reasons (we use 1K = 1000 bytes for all the other packages).

PR #24, opened 2016-02-06
WEB: Use German release titles in news

rootfather:
BS2.5 and 1.8.0 testing announcements now use the German release titles instead of the English ones.


criezy:
Merged and updated the web site. These changes should now be live.
Thank you.

PR #25, opened 2016-02-06
WEB: Fix and improve downloads section on German website

rootfather:
This PR fixes the superfluous {release} that's currently displayed on the German downloads site (thanks to @wjp). download.de.xml now uses the formal mode of addressing instead of the informal one.


sev-:
No, I do not want to have downloads.de.xml. It must be addressed via #variable# substitution. Otherwise we end up with updating download list for every language, which is not very convenient.

rootfather:
Okay, currently not all strings that can be translated are merged into
lang.ini.

I'll have a look on how to merge the remaining strings to the lang.ini file
tonight.

criezy:
Yes, I started to add variable substitution to it. This was not finished and thus I kept the downloads.de.xml but it should really be removed.

The issue I had was that variable substitution didn't see to work for text in <name> and <extra_info> tags in the file blocks.

sev-:
probably I'll need to enhance the parcer a bit

rootfather:
I currently see two options to get this resolved:

1. keep downloads.de.xml as temporary solution and merge this PR to resolve the {release} bug on the German website. This would make the translation pretty much complete, but we would have to maintain two files. Quite a bad idea, now that we are in the middle of a release cycle.
2. delete downloads.de.xml and only keep the changes I made to the lang.de.ini. On the positive side, this would make future maintenance much easier, but we would loose some translations. However, this would make "hunting" for missing strings easier too.

If you ask me, I would prefer option 2. 

What do you think? Or should we just wait with this until the parser is enhanced as sev- proposed?

criezy:
I would say go for option 2 as well.
You can update this pull request, or if you prefer I can just merge the changes in lang.de.ini and delete the downloads.de.xml file myself.

rootfather:
I deleted downloads.de.xml, so we are ready for merging this PR I guess.

criezy:
Thank you. Merged.

PR #26, opened 2016-02-22
Italian translation

tobiatesan:


criezy:
Did you omit on purpose the locale and date format strings at the start of the lang.it.ini? Don't you want to translate dates to Italian?

It is also missing the some of the strings (for example all the strings from downloads.xml). It looks like you based your translation on an older version of the english lang.ini?

tobiatesan:
I did. Thanks for catching it.

sev-:
Any update?

tobiatesan:
@criezy I've integrated the changes from cbda272, how does this look?

sev-:
Merging, though the "new style paths" are present in menu.xml too. I'll take care of it.

PR #28, opened 2016-02-22
WEB: Update FAQ

criezy:
Some of the answers in our FAQ on the web site are outdated (for example on making new games or on AGS being closed source). Here are suggestions to improve the FAQ. Feel free to disagree with some of my changes or propose better formulations.

### DIFF
--- a/data/faq-xml.xml
+++ b/data/faq-xml.xml
@@ -88,7 +88,7 @@
 		</simpara>
 		<simpara>
 		ScummVM runs on a wide variety of platforms, including Win32 (Windows
-		95/98/ME/NT/2000/XP), Linux i386 and PPC, Mac OS X, many smart phones
+		95/98/ME/NT/2000/XP/7/8/10), Linux i386 and PPC, OS X, many smart phones
 		and game consoles and more. It is also part of various BSD ports
 		collections and included in Debian testing and unstable.
 		</simpara>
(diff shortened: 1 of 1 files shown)


rootfather:
The changes you made to the FAQ are very good if you ask me.

I have a question regarding the supported systems: While working on the German translation of the FAQ (slowly but steady :)) I noticed that Windows 95 is listed as supported system. Is this still true or an outdated information? I don't have a Windows 95 CD anymore to test this on a virtual machine.

lordhoto:
Win95 should still work.

lordhoto:
The changes look good to me. I added a nitpicking comment to one place though :-).

rootfather:
Thanks for the information. I should have a Windows 95 or 98 disk laying
around somewhere around here. I'll give it a try just to check it.

But anyway, we should add Windows versions after XP until Windows 10 to the
FAQ.

criezy:
I made suggested by lordhoto to use 'Macintosh' rather than 'mac' (it actually appeared in a few other places in the FAQ and I changed it everywhere).

In the same commit I also changed 'Mac OS X' into 'OS X' since this is the official name now. That one might be a bit controversial though since this is in the list of platforms we support, and we support old versions that are officially named 'Mac OS X'. But I didn't think it was worse listing both.

criezy:
I now updated the list of supported Windows version to list the recent ones. I have to wonder though if we should really list all the versions. We don't do that for other OS's. Or we could just write 'all Windows versions from Windows 95 up to Windows 10'.

criezy:
If there is no objections I will squash the three commits and push to origin/master on Sunday evening.

data/faq-xml.xml
```diff
@@ -527,7 +553,9 @@
 		These are not full games but reencoded cutscene (video) packs. To run the games you
 		still need the original disks. Originally cutscenes were encoded in a proprietary
 		format but Revolution Software kindly permitted us to reencode them into
-		DXA and distribute them freely on our site.
+		DXA and distribute them freely on our site. Note however that recent ScummVM versions
```
RichieSams: One small suggestion for grammer:

"Note, however, that recent ScummVM versions support using the original cutscene files directly. Therefore, you don't have to use the ones from our site anymore."

criezy:
I have now pushed the commit to master.

PR #29, opened 2016-02-28
Fix buildbot translation

rootfather:
This PR removes the hardcoded "generated by our buildbot" string in downloads.xml and replaces it with a placeholder called "DownloadsDailyLink4". This allows translations that require additional stuff after the word "buildbot". This is required for the German translation and maybe for some other languages too.

Thanks to WindlePoons from IRC for pointing that out.

I hope I used the string placeholder the right way.


criezy:
Thanks. I pushed the commit to our repository.

PR #41, opened 2016-03-08
WEB: Translate introHeaderContentP5 to Italian

tobiatesan:
I found this one considerably difficult to render in Italian while preserving meaning *and* tone, but I think I nailed it.


criezy:
Thanks. Merged.

PR #45, opened 2016-03-08
Ita news

tobiatesan:
Given the nature of 20160119 I would please ask somebody, especially @Strangerke, to double-check it (with the Italian I believe he knows a bit _and_ Google Translate) and make sure it is exactly what he wanted to say.

Please ask if unsure :)

Thanks.


sev-:
Thanks

PR #47, opened 2016-03-11
WEB: German translation cleanup

raziel-:
- Cleanup
- Consistency
- Typos
- some stuff we can fight over :-)


rootfather:
I agree with all the changes - no objections against merging this from my side. Thanks for the corrections!

sev-:
Thanks

PR #49, opened 2016-04-01
WEB: Add John Carmack April Fools' news

tobiatesan:
Just an idea - I have had this image on my disk for a while now and it makes for a very... er, nice clickthrough.
If you like it tweak it as you please :)


sev-:
No, the screenshot is good fore ResidualVM, not for us. Rejecting.

PR #53, opened 2016-04-23
Move language button to top of screen, mobile fixes

### LINKED ISSUE #52: Language menu not visible on iOS and Android
Today SimSaw told me he could not see the language menu on his android phone while he could in the past. I don't have an android device to check, but I checked on my iPad and I have the same issue.

Using the WebInspector to highlight it I could determine that the language menu is not at the correct location: it is around the Forums links on the main menu (but is not visible):

webinspector_mobilesafari

salty-horse:
1. Fixed #52 by using a larger `min-width` for the page.
My earlier change made it disappear, but even before that, the language button was hovering over the logo.

2. The button was unusable on mobile. Since mobile doesn't have "hover", you have to click on it. Clicking on it reloaded the main page since it was a `<a href="">`. I fixed this by changing the `<a>` to a `<span>` and prettied it up by changing the cursor to a pointer. I also took the opportunity to remove some unused CSS, and vertically center the "language book" image in the tab.

### DIFF
--- a/css/lang-menu.css
+++ b/css/lang-menu.css
@@ -19,14 +19,6 @@ code.text-badge {
 	padding: 0 0px;
 }
 
-#langmenu a {
-	position: relative;
-	top: 4px;
-	color: #000000;
-	text-decoration: none;
-	font-size: 9pt;
-}
-
 #langmenu > li {
 	float: left;
 	height: 29px;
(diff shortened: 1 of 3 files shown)


salty-horse:
Deprecated by #54, which moves language switching to the sidebar.

salty-horse:
In addition to the changes in the first comment, I also moved the language button to the top of the screen. Please review :)

I tested the mobile behavior in Chrome mobile simulation. It has a small bug where if you click on the language button while the screen is completely zoomed out, it seems to auto-click the first option (as if it's registering two clicks). I wonder if it also happens on actual mobile devices.

salty-horse:
I also removed the spacing between language entries. Notice how moving the cursor over the entries always highlights one of them, where before there was some white space. I like it better, hope you agree.

sev-:
Nice, thanks!

PR #54, opened 2016-04-23
WEB: Move language switching to the sidebar

salty-horse:
This patch moves the language switcher from a tab at the top of the page, to a sidebar (requires some scrolling to reach).

The old language switcher position was odd, right between the logo and "Script Creation..." text.


sev-:
No, I honestly not in favour of this direction. There are several reasons for that:
* The page becomes extra long. Thus, some short ones like Contact Us, look weird.
* Then, because of the reason above, the tentacle is now overlapping the menu on those short pages.
* The next thing is that now with translations the menu could get very long. Wikipedia renders languages in a small font for avoiding that
* Also, now we're ruining the concept of the menu, as some of the items do not lead anywhere but just change the language
* Another thing, is that with this rendering it is not immediately noticeable that there is a way to switch the language. Many users may read the whole page before seeing the switcher.
* And finally, our Facebook/Twitter buttons, as well as other, become even less noticeable than they're now.

Initially, I was thinking of putting the language drop-down menu on the very top, e.g. y coordinate is 0, and have the bottom part of it rounded. That is, putting the current "closed" rendering upside down. Then the language switcher will be prominent, not obscuring or damaging any of existing design, and do not clutter the page.

If you could help me with achieving that with CSS, that would be fantastic

salty-horse:
I'll try your suggestion of putting it at the top of the screen. Closing this PR.

PR #55, opened 2016-05-04
WEB: Center "Language" button

Tkachov:
Simply centered button.

This solution uses `transform` property, which is CSS3, but should be supported quite well. There are browser-specific prefix-versions of this property, which are supported by older versions of browsers, but probably still not all of them.

It looks like that: Centered button

@salty-horse says he assumed it's supposed to be centered between the logo and the characters instead of the whole page.

### DIFF
--- a/css/lang-menu.css
+++ b/css/lang-menu.css
@@ -1,8 +1,8 @@
 #lang-wrapper {
-	margin-right: 10px;
-	float: right;
-	position: relative;
-	top: 0px;
+	width: 100%;
+	position: absolute;
+	left: 0;
+	top: 0;
 }
 
 code.text-badge {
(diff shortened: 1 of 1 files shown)


salty-horse:
On Chrome, the drop down list is obscured by the "What is ScummVM" box.

Tkachov:
Well, `<ul>` has `z-index: 1000;`, so it's kind of strange, but `z-index: 10;` (which I added for `#langmenu`) worked for me in my Firefox. May be we should use 1000 there as well.

Tkachov:
Well, adding `z-index` to `#lang-wrapper` instead of `#langmenu` fixed the situation (at least in my mobile Chrome).

I also have the other solution which doesn't require neither changing `z-index` nor using `transform`, but I don't like it as we have to have fixed width for that "Language" button.

bluegr:
Can't you just do:
margin: 0 auto;
text-align: center;

The parent element would have to be 100% in width, with a relative position.

That seems easier than your solution... Is there a reason why it wouldn't work?

Tkachov:
That's the "other" solution I mentioned. It requires button element (`#lang-wrapper`) to have fixed width (like `width: 100pt`). As I don't really know which width it should be, I used the solution which would work with any width.

Plus, the parent element has `float: right` and doesn't have `width: 100%;`, but of course that's not difficult to remote the former and add the latter.

PR #58, opened 2016-06-21
Fix typo

AMDmi3:
https://en.wiktionary.org/wiki/совместимость


AMDmi3:
Cast @sev-

sev-:
Thanks for fixing. Merged.

PR #61, opened 2016-10-29
WEB: Remove links to sf.net project home

rootfather:
Now that we moved everything (downloads, ticket system, mailing lists...) away from sf.net, I think it makes sense to remove the last remaining links to the sf.net project home.

### DIFF
--- a/data/menus.de.xml
+++ b/data/menus.de.xml
@@ -56,10 +56,6 @@
 		<name>Entwicklung</name>
 		<class>menu-sf</class>
 		<link>
-			<name>Projektseite auf SF.net</name>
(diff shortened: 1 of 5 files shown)


criezy:
Maybe rather than removing the link we could replace it with a link to the project page on GitHub (https://github.com/scummvm)?
Although we do have a link to the source code below (pointing to https://github.com/scummvm/scummvm), so that might be redundant.

rootfather:
I consider the scummvm.org to be the new project home itself. We link to everything from here (self-hosted bug reports, the code repository, downloads, the mailing lists), so I really don't see that we need sf.net anymore.

criezy:
Thank you.
I agree with your reasoning and just merged this change.

PR #64, opened 2017-03-07
WEB: Add PSP2 (Vita) buildbot icon

rsn8887:
I would like some feedback if I am doing this right. I am trying to add a buildbot for the Vita version, and this is the missing icon.


sev-:
The icon is missing shadow. See the following page on how to make it: http://wiki.scummvm.org/index.php/Screenshots#How_to_make_game_icon

rsn8887:
Shadow added.

PR #65, opened 2017-07-04
Adding Plumbers Don't Wear Ties to the compatibility list

Thunderforge:
Added to ScummVM by commit e4ab357 on May 2.


Thunderforge:
What still needs to be done for this to be merged?

digitall:
@Thunderforge : I don't think any further work is required and @tobiatesan has reviewed it and it looks fairly trivial to me, so will merge this and let the ScummVM web team deal with any further minor changes required.

PR #70, opened 2017-12-22
WEB: Remove links to the old Feature Tracker

Mataniko:
The menus still linked to the old feature/bug tracker on sourceforge.


rootfather:
Thanks Mataniko. Shouldn't we redirect the menu entries for the feature requests to the new bugtracker instead of removing them entirely?

sev-:
Thanks!

PR #72, opened 2017-12-23
WEB: Always redirect requests to https on the main site

Mataniko:


sev-:
Why? The whole site is read-only.

hpvb:
You get ranked higher on Google if you're https only

rootfather:
I personally think that everything in the Web should be HTTPS. I also think that modern browsers will start showing up warning messages when they hit unencrypted content. In general, there should be no downside in switching over to HTTPS even on read-only sites like the main page.

Mataniko:
While not required on the main site, there's no real downside either since it's already set up for HTTPS. 

The real case for this is that the wiki and forums need to be switched to HTTPS as they both have a login form transmitting passwords in the clear - see https://bugs.scummvm.org/ticket/10375. It doesn't make sense having the main site HTTP and then direct people to HTTPS subdomains

csnover:
This absolutely should be done across all sites. Unencrypted HTTP should never be used any more (in fact it is impossible to use HTTP/2 without encryption in any browser which supports it).

rootfather:
This should be another valid reason to start shipping everything ScummVM/ResidualVM related downloads via HTTPS:

https://citizenlab.ca/2018/03/bad-traffic-sandvines-packetlogic-devices-deploy-government-spyware-turkey-syria/

tl;dr: Providers are redirecting HTTP download requests for popular open source software to their 'own' packages infected with malware.

sev-:
Thanks, merging.

PR #75, opened 2018-05-27
WEB: Add screenshots for Rama

dafioram:
Screenshots and thumb provided by dafioram
and windlepoons in  #Trac10264.


lotharsm:
Please note that the correct syntax for the filenames with only one digit is just "x" instead of "0x". This should be fixed before I can merge this.

sev-:
Nice, keep'em comin'

PR #76, opened 2018-05-27
WEB: Add screenshots for Titanic

dafioram:
Screenshots and thumb provided by dafioram
and windlepoons in #Trac10030.


lotharsm:
Please note that the correct syntax for the filenames with only one digit is just "x" instead of "0x". This should be fixed before I can merge this.

dafioram:
Many screenshots do not follow that convention, like Full Pipe.

sev-:
And that is correct, as there is format specified, %02d.

There is a different problem, however. The category icon does not have a shadow, as it is required.

The instructions on how to make that icon could be found here:
   http://wiki.scummvm.org/index.php/Screenshots#How_to_make_game_icon

Same problem is with the recent Lighthouse category icon, but it was already merged....

dafioram:
Shadowing added to icon.

sev-:
Great, merging.

PR #84, opened 2018-07-04
WEB: List game titles for screenshots in alphabetical order

dafioram:
I used the convention that if a game title starts with a "The"
then the next word is used for determining order.

### DIFF
--- a/data/screenshots.xml
+++ b/data/screenshots.xml
@@ -5,57 +5,91 @@
 		<title>LucasArts games</title>
 		<category>lec</category>
 		<game>
-			<name>Maniac Mansion</name>
-			<category>maniac</category>
+			<name>The Curse of Monkey Island</name>
+			<category>comi</category>
 			<image>
-				<file>lec/maniac/scummvm_0_0_#n#</file>
+				<file>lec/comi/scummvm_0_11_#n#</file>
+				<range from="0" to="2" format="%d" />
+				<caption>The Curse of Monkey Island</caption>
+			</image>
+			<image>
+				<file>lec/comi/comi-#n#</file>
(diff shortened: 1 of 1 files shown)


lotharsm:
Thanks for your work!

Just my personal opinion, I would prefer to have a naming scheme like "Curse of Monkey Island, The" instead, have you considered that?

criezy:
Thank you. Currently it looks like some games are in alphabetical order (for example the Revolution Software games and SCI games (almost - one game is misplaced) while SCUMM games for example appear to be sorted from older to newer. Some consistency would indeed be nice. 

Interestingly the Compatibility page is similarly inconsistently sorted and seems to be using mostly alphabetical order, but not for SCUMM games. Maybe that one should be changed as well?

Alphabetical order is fine for me. I think it makes it easier to find a game as I do now my alphabet while I don't know the release date of each game.

Also personally I don't like the naming scheme proposed by @lotharsm as it breaks the natural reading of a text and makes it harder to read.

Mataniko:
We can also simply call `sort($games)` in https://github.com/scummvm/scummvm-web/blob/c4c59152150042ef450827d464df8c0b91d8617d/include/Models/ScreenshotsModel.php#L25 instead of manually sorting the XML every time which will be prone to errors.

I don't have a personal preference for games starting with "The" and how they're sorted, but we should be consistent.

dafioram:
That works for me. Is it also possible to sort the game titles on the compatibility page also?

Mataniko:
@dafioram Should work the same way, look in CompatibilityModel.php

dafioram:
Having it automat

Issue #80, opened 2018-07-01
Unusable garbage on Chromium 66


hamakei:
Looks like some badly formatted mobile site from 2003.

![untitled](https://user-images.githubusercontent.com/30203431/42138398-e870762c-7d74-11e8-8651-3113d5a053d5.png)

Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/66.0.3359.181 Chrome/66.0.3359.181 Safari/537.36

hamakei:
Makes no difference. Even browsing in private mode (with no cache or cookies) it looks like that.

tsoliman:
Tested with Chromium 67 on Linux (Debian 9 64bit) and it works fine.
(Fresh install in a VM)

tsoliman:
@hamakei is an extension blocking something?
Are there any errors in the JS console? (ctrl-shift-j)
Any non-200s in the Network tab?

hamakei:
Ah. There is a javascript error...
"Failed to load resource: net::ERR_BLOCKED_BY_CLIENT" - pixel.gif from paypalobjects.com

Mataniko:
@hamakei I've tried to reproduce using Chromium 66 on Ubuntu with uBlock Origin. 

I see pixel.gif being blocked, but the rest of the site loads fine. Do you have any other addons installed that might be interfering? Any other information from the console that you can share?

metakid:
I see the same thing on android, on a samsung S6 in chrome. The responsive is not responding. It does sort of make you feel like you were pulled into the past by some kind of space vortex thing. Which I guess is sort of the point of this project. :|

Mataniko:
Can you consistently reproduce? It feels like you're getting a cached version. 

If possible, use Chrome remote debugger to capture a network log and send

bonki:
I can't reproduce this either. Tested with Chromium 37, Chrome 65 and 67 on Windows, Linux and Android and different versions of uBlock Origin.

PR #85, opened 2018-07-07
Use SCSS instead of CSS

Mataniko:
Hi everyone,
Another fairly big change that will be the foundation to future updates.

I've started breaking apart all of our CSS into reusable components. There are multiple advantages from the move.

- The compiler will always generate syntactically correct CSS.
- Easier to maintain, adjust and update
- Theme support, there's still some work needed but this would be easily transferable to our sister project, ResidualVM 
- Component based structure
- Outputs a single minimized CSS file.

There are a couple of other subtle changes that were done to get this working, including using Composer which will simplify deployment, and opens up the other PRs I have lined up.


Mataniko:
Note there are no actual design changes to the website, it's all back-end improvements.

include/Pages/CompatibilityPage.php
```diff
@@ -29,7 +29,7 @@ public function __construct() {
 			'untested' => 'pctU',
 			'broken' => 'pct0',
 			'bugged' => 'pct50',
-			'good' => 'pct90',
+			'good' => 'pct75',
```
sev-: Why? Good is 90% intentionally.

scss/vendor/_normalize.scss
```diff
@@ -0,0 +1,342 @@
...
+  }
+  
+  /**
+   * Add the correct display in IE 10.
+   */
+  
+  [hidden] {
+    display: none;
+  }
+
```
sev-: No endline

include/Pages/CompatibilityPage.php
Mataniko: I generate the CSS colors from red to green programmatically. The 75% value is what the color was in .pct90.

If we were to use the actual .pct90 value, it would be almost indistinguishable from 'excellent'

Mataniko:
I'd like to get this merged this weekend, is everyone else OK with the change? It should also resolve #86 and #80 because it's most likely a caching issue.

lotharsm:
You have a "Go!" from my side.

criezy:
I basically have no idea what this all means. I had never even heard of SCSS before you opened this PR. You list some benefits from this change, but possible drawbacks are not clear to me. For example:
-  Does it have any impact of the range of supported browsers, or would any browser that correctly displays the website as it is now (after your previous changes) still display it correctly after the changes in this PR?
- Does it have any impact on the performances (e.g. how long it takes the website to display on a low end device)?

You seem to imply that those are just internal changes that will have no impact for the end user. If that is the case I have no objection to the change.

Mataniko:
@ciezy:
Great questions.

SCSS is compiled into CSS during site-update, so the output is the same as the CSS we have today. I'm working on integrating with a CSS postprocessor that will increase support to older browsers by automatically appending vendor prefixes to any future features that are more experimental (we use none at the moment.

It should improve performance since instead of serving multiple CSS files per page, everything is now built into one, minified CSS. The biggest impact on performance when serving websites is the number of network requests needed to render content.

criezy:
No objections from me then. This sounds like a good update.

PR #87, opened 2018-07-21
WEB: Redo all category icons for games and platforms and move them to a spritesheet

Mataniko:
Both games and platforms icons we use are pretty bad. There are also enough icons that loading /screenshots or /platforms is slow because of how many connections are needed to download the images.

This PR includes:

- Brand new or updated icons for every single category & platform
- Updates site-update to generate a spritesheet automatically
- Updated CSS/HTML to use spritesheets instead of individual images
- Moved all category images to subfolders
- Increases the icon size slightly from 29px to 32px

Some icons are still not great, but those can be addressed individually.


Mataniko:
Note that we will need to install Glue via pip first before merging!

Mataniko:
**Old platforms sprite:**
platforms

**New platforms sprite:**
platforms-new

**Old games sprite:**
games

**New games sprite:**
games-new

Mataniko:
@sev- Test website is up at https://scummvm.herokuapp.com/

sev-:
No, it is significantly uglier than I even anticipated. The shadows are ugly (the previous ones made it look like the icons are convex, and the current one is like they're floating). The icons are not centered vertically, not balanced, e.g. vary in size relative to each other, and are disproportionally big in comparison to the text.

Needs more work to address the above.

screen shot 2018-07-22 at 17 37 16

Mataniko:
@sev- I've done some re-balancing in the images:
image

What do you think?

Mataniko:
Updated all the icons, preview at https://scummvm.herokuapp.com

Mataniko:
games 2x
platforms 2x

criezy:
The new icons are quite nice I think. The are still bigger than what we currently have on the live web site, but I think it is better that way. The game icons on the live website are too small to recognize anything (at least on my 2560x1600 30'' screen).

I think the new platform icons look nice. I also like most of the new game icons, although the one for Lure of the Temptress for example would benefit from having less details (or maybe more contrast?) I think. The size and alignment is also more consistent now compared to your first version.

lotharsm:
I like the new icons too, good work!

The font size could be a tad smaller in my opinion.

Issue #89, opened 2018-07-26
Add cookie consent banner


Mataniko:
We use Google Analytics to track website statistics. In order to comply with various EU regulations the website needs to show a cookie consent banner to the user.

PR #90, opened 2018-08-23
Add a cookie consent form for Google Analytics

Mataniko:
To comply with EU regulations, we must ask customers if they're willing to let us use tracking cookies. This PR adds a floating form on the bottom that requires the user to consent to in order to enable google analytics.  

There doesn't seem to be a way to use GA without it storing a cookie, so if the user doesn't provide consent, the script never even loads.

This addresses #89 

image

### DIFF
--- a/lang/lang.ini
+++ b/lang/lang.ini
@@ -416,3 +416,8 @@ Currently both subprojects are still in a state only suitable for
 developers... so if you can't compile the code yourself, then these are
 not really ready for you.
 """
+
+# cookie.tpl
+cookieText = "We use cookies to enhance content and analyze information on site performance and usage."
+cookieAccept = "Allow Cookies"
+cookieDecline = "Refuse Cookies"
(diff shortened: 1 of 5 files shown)


lotharsm:
I wonder if it is possible to translate the cookie consent notice to other languages to be fully GDPR compliant?

lotharsm:
Thanks!

Mataniko:
@lotharsm I'm not sure this was ready to merge. I'd like @sev- to take a look since we had a discussion around how non-consent and non-response should be handled.

lotharsm:
Damn, seems that I was too fast then... what to do know?


________________________________
Von: Matan Bareket <notifications@github.com>
Gesendet: Montag, August 27, 2018 7:56 PM
An: scummvm/scummvm-web
Cc: Lothar Serra Mari; Mention
Betreff: Re: [scummvm/scummvm-web] Add a cookie consent form for Google Analytics (#90)


@lotharsm<https://github.com/lotharsm> I'm not sure this was ready to merge. I'd like @sev-<https://github.com/sev-> to take a look since we had a discussion around how non-consent and non-response should be handled.

PR #93, opened 2018-11-17
WEB: Update debian package version

Fil0sOFF:
v2.0.0 is already in buster and sid

### DIFF
--- a/include/config.inc.php
+++ b/include/config.inc.php
@@ -2,7 +2,7 @@
 /* Current version. */
 define('RELEASE', '2.0.0');
 define('RELEASE_TOOLS', '2.0.0');
-define('RELEASE_DEBIAN', '1.9.0');
+define('RELEASE_DEBIAN', '2.0.0');
 /* Version when the percentages on the compat page were removed */
 define('COMPAT_LAYOUT_CHANGE', '1.7.0');
 


digitall:
@Fil0sOFF : Thanks for the nudge.

@Mataniko : I did a search of the web code as the Debian release packages on the downloads page are already v2.0.0 and I wasn't sure what this was for. It is just to put the note on the downloads page near the top "1.9.0 is also apt-get'able from Debian unstable (sid)." ...

I think we might consider removing that along with the downloadsBinariesNote3 and release_debian / RELEASE_DEBIAN references since it should be obvious that users should use the versions from their operating system app stores / repositories unless they are badly outdated or they wish to use the latest stable / development snapshot.

digitall:
I think this was added as Debian (stable, especially) and thus Debian derived distributions such as Ubuntu and Raspbian tended to be on some very outdated versions i.e. several years behind.

However, I would suggest a more general and less version specific message might be better i.e. "The following downloads are provided for the latest stable version by the ScummVM Team. While they are the official installers for various operating systems, users should preferentially use their operating system repositories or app stores to install the latest ScummVM unless these are providing outdated versions"

Might be worth a post to scummvm-devel?

digitall:
But as a first instance, fine to just merge this and nudge RELEASE_DEBIAN up to date.

Fil0sOFF:
@lotharsm it's done.

lotharsm:
Thanks! Merging then.

PR #95, opened 2019-03-10
WEB: Update logo, favicons and background colors

Mataniko:
For the new year I've been exploring modernizing the branding. Our current logo was designed almost 15 years ago, and looks a little out of place this day and age. We also heavily rely on simple gradients that sometimes that don't look good in some of the resolutions we're on.

While still keeping the same look and feel, the proposed logo is flatter, but still contains depth. Colors have been adjusted as well.

Old Logo:
scummvm_logo

New Logo:
scummvm_logo

Old Icon:
scummvm_icon

New Icon:
scummvm_icon

Old Tools Icon:
scummvm_tools_icon

New Tools Icon: 
scummvm_tools_icon_512

Website Screenshot:
Screenshot_2019-03-10 ScummVM


peterkohaut:
I love it!
But those gradients in news headers (red) and section headers (yellow) now feels a bit dated and out of the new design

lotharsm:
Wow, impressive job!

Being a fan of a "clean and modern" artstyle, I really like this new look and feel. However, I ageed with @peterkohaut that the gradients should be removed as well.

peterkohaut:
I hacked in a quick prototype without gradients (+some border adjustments) just for preview

Screenshot_2019-03-10 ScummVM Home


[main.css.patch.txt](https://github.com/scummvm/scummvm-web/files/2950067/main.css.patch.txt)

Mataniko:
@peterkohaut Thanks for the quick patch. I've been playing both with removing the gradients and a redesign as a whole but wanted to keep this PR self contained. This is definitely something that I'll be adding to the website.

@criezy That's correct - I have the media package updated as well.

lotharsm:
Just in case we want to use the new coloring scheme for the GUI as well, I made a really quick and really dirty hack for the GUI theme files of the "Modern Theme"... Just as a very first impression on how things *might* look in the future.

2019_mockup

Mataniko:
Another one with the same orange used in the website background: 
image

I don't think the orange works with the red buttons, that's where the yellow part of the gradient really helped

Mataniko:
I'd like to also provide another green shade as an alternative, greener but not meaner.

Proposed
scummvm_logo_large

Greener
scummvm_logo2

bluegr:
Nice work! I like the new icons a lot :)

The greener logo looks a bit more lively. There's no major difference, but it's slightly better.

I'm a bit undecided on the orange background. Perhaps some other color would also work?

digitall:
Overall, this looks reasonable, though am a bit personally ambivalent on the "modern" tendency to make UI elements flat as it can result in bad contrast or unclear what elements are interactive.

However that orange background really looks awful to my eye... I assume this is the average of the current dark orange to yellow gradient used on the current Modern Theme. I would suggest either the yellow or the dark orange of the current theme, but not the midpoint as it result in something which really grates. Though I admit this may just be my personal dislike for pastels.

Mataniko:
@digitall I agree, I didn't want to go too flat, so I introduced depth on Scumm part plus shadow across the board to give the whole thing some weight. I played around with adding depth to the VM part, but it became very busy.

The problem with too much depth that at low resolutions you lose fidelity. This hopes to achieve balance between "too flat" and "too deep".

Tkachov:
Putting my two cents in, in 2017 I sketched some redesign for the site and app, and made a prototype theme for old design with the color scheme I used. Changing logo to a new one, it would look like this:

prototype_2019_touch

The colors might be too bright and too different to current ones, but in my 2017 opinion they fit better than orange/red ones.

Mataniko:
> Putting my two cents in, in 2017 I sketched some redesign for the site and app, and made a prototype theme for old design with the color scheme I used. Changing logo to a new one, it would look like this:
> 
> prototype_2019_touch
> 
> The colors might be too bright and too different to current ones, but in my 2017 opinion they fit better than orange/red ones.

I don't love the colors, but I think this could be a separate discussion worth having.

Mataniko:
I updated the colors following the feedback in the thread. Made the Green greener and adjusted the orange hue. The website would look like this:
image

While the app would look like this:
image

sev-:
To me, this green looks too acid.

Mataniko:
scummvm_shades

Various shades of greens

bonki:
I'll let it sink in a little but right now I think I prefer the first shade of green in your last image.
Is there a particular reason why the outline is of a much lighter grey than in our current logo (where it looks black)?
Also, is it just me or is the "VM" part a tad bigger than currently?

Thank you for working on this!

Mataniko:
@bonki I think the black makes it look too flat, since it becomes more of an outline and less of a layer. Here's a preview: 
image

The VM part hasn't been changed, nothing has been resized.

bonki:
Call me whatever but I like that \*shrug\*.

lotharsm:
Personally, I prefer the grey-ish border - I think it's a nice touch. Color-wise, my favourite is the first one, it integrates well with the orange tone you've chosen.

criezy:
In term of green I am not sure which one I prefer, but I know it is neither the second one (it looks meh) nor the last one (too aggressive). I think it might be between the 1st and the 5th ones. And I am with bonki for the contour: I actually prefer the darker one. I thinks it gives it more definition.

Mataniko:
Seems like we are converging towards a consensus, or at least close enough for one. So here we have it:

1st green, gray and black backdrop on orange:
image

Tkachov:
I guess you don't need my opinion, but here's one anyway (=

IMO №1 and №2 are "unhealthy" green. The desease-kind-of-green.

№3 and №4 are almost the same, and they are somewhat of healthy grass green, peaceful and not too bright.

№5 and №6 are brighter variations, and №6 is looks acid. №5 is something in between №3-4 and №6 -- it's bright, but it's still healthy green.

Regarding outline: though I thought grey would be better, on the comparison the one with black stands out, so if that's the point, black should be used

Finally, I really don't like that not-so-orange tone. It couldn't be bright as it's a background, but this semi-dark tone is awful. That's why in 2017 I tried to find another color, and decided to give monochromatic a try, resulting in that pastel yogurtish light green background. The complementary color is pink, which is too bright for background, and looks strange in a lighter/darker shades.

I experimented a little bit and it looks like purple is a funny match (kinda Mysterio-ish or Day of the Tentacle vibe), but it's only OK if you want to make a dark theme, which I guess is not what ScummVM needs. Two examples:

purple fw
purple2 fw

Mataniko:
I can always bring back the old theme :)
image

lotharsm:
aaaand I have a new favorite: @Tkachov 's No. 1. @Mataniko, hm, the new-old theme doesn't look bad either...

lotharsm:
The more I think about and look at the purple themes, the more I start prefering it to the current orange. To be honest, the "new old" website theme looks great to me and Tkachov's first color combination is a good starting point for the GUI.

What do you guys think?

Mataniko:
This isn't the right place to discuss a change of color palette. Let's either close the loop on the logo or abandon the PR.

lotharsm:
Okay then, let's get back to the logo itself then. The last proposal from @Mataniko including the grey borders look fine to me.

Mataniko:
Thank you everyone for the feedback. It's been a mixed bag between the different colors and looks but at the end I think the process helped and improve the end results.

I'd like to present:
og-image

image


Notes:
- The green shade has been changed to match the highlights on Purple Tentacle. This doesn't have the "sickly" green look, not too bright and fits perfectly with our history.
- The border was too gray, this is somewhere closer to black, with enough lightness in it to add depth.
- The colors are in the CYMK color space, meaning they will print well.
- The orange color is now bolder and contrasts better with the logo.
- Gradients have left the building

Overall, if there are no major objections, I'm looking forward to merge this over the next couple of days.

waltervn:
Looks great. Nice work!

Walter

Van: Matan Bareket
Verzonden: dinsdag 26 maart 2019 07:11
Aan: scummvm/scummvm-web
CC: Subscribed
Onderwerp: Re: [scummvm/scummvm-web] WEB: Update logo, favicons and backgroundcolors (#95)

Thank you everyone for the feedback. It's been a mixed bag between the different colors and looks but at the end I think the process helped and improve the end results.
I'd like to present:


Notes:
• The green shade has been changed to match the highlights on Purple Tentacle. This doesn't have the "sickly" green look, not too bright and fits perfectly with our history.
• The border was too gray, this is somewhere closer to black, with enough lightness in it to add depth.
• The colors are in the CYMK color space, meaning they will print well.
• The orange color is now bolder and contrasts better with the logo.
• Gradients have left the building
Overall, if there are no major objections, I'm looking forward to merge this over the next couple of days.

criezy:
I like it!
Thank you for your work.

sev-:
I like it too!

PR #96, opened 2019-03-30
WEB: Remove the reference to Debian unstable; use generic wording instead

lotharsm:
Since current versions of ScummVM are not only available in Debian sid, but also in many more bleeding edge Linux distributions like Arch, Manjaro, Fedora, openSUSE Tumbleweed, current Ubuntu and many more, I propose to remove the reference to Debian sid on the download page and use a more generic wording instead:

"ScummVM {release} is also available in the software repositories of many GNU/Linux distributions."

Please let me hear your thoughts.

### DIFF
--- a/lang/i18n/lang.en.json
+++ b/lang/i18n/lang.en.json
@@ -75,7 +75,7 @@
     "downloadsBinaries": "{release} Release binaries",
     "downloadsBinariesNote1": "For a list of changes since the previous version,",
     "downloadsBinariesNote2": "read the release notes",
-    "downloadsBinariesNote3": "{release_debian} is also apt-get'able from Debian unstable (sid).",
+    "downloadsBinariesNote3": "ScummVM {release} is also available in the software repositories of many GNU/Linux distributions.",
     "downloadsSourceCode": "{release} Source Code",
     "downloadsTools": "{release_tools} Tools",
     "downloadsOldBinaries": "Older versions (unsupported)",


tsoliman:
How about we say "Debian Sid and also many other distributions"

(My rationale is that with your proposed wording, Debian stable won't have it, therefore Debian as a whole is implied to be unpopular)

lotharsm:
@tsoliman, thanks for your feedback!

> How about we say "Debian Sid and also many other distributions"
Sure, that's also an option. I just wonder why we mention Debian sid explicitly here?

We currently have a .deb package for the current Debian stable, and the next Debian release which is already in a frozen state will ship with ScummVM 2.0.0 by default.

lotharsm:
@tsoliman, thanks for your feedback!

> How about we say "Debian Sid and also many other distributions"
Sure, that's also an option. I just wonder why we mention Debian sid explicitly here?

We currently have a .deb package for the current Debian stable, and the next Debian release which is already in a frozen state will ship with ScummVM 2.0.0 by default.

criezy:
Why would the proposed wording imply that Debian stable won't have it? Is that not included in "many GNU/Linux distributions"? I don't really see how it can give the impression that Debian is unpopular.

To me the proposed change seems good.

tsoliman:
> Why would the proposed wording imply that Debian stable won't have it? 

Because Debian stable almost never has the latest anything, hence the pointer to sid.

lotharsm:
@tsoliman, thanks for the feedback!

If there are no further objections, I will merge this tomorrow.

lotharsm:
Hm. @mataniko any idea why the variable substitution for {release} doesn’t work here?

tsoliman:
I am not a PHP expert. 

https://github.com/scummvm/scummvm-web/blob/d671d448644b0952f8493efc3658c4940671647f/include/Pages/DownloadsPage.php#L70

My guess is that `'release_debian' => RELEASE_DEBIAN,` on that line should be changed to `'release' => RELEASE,` since release_debian didn't seem to be used anymore.

At least that template var should be added.

lotharsm:
Thanks, good catch. I assumed the release variable was available globally, so adding it should fix it :)

PR #97, opened 2019-04-25
WEB: Add redirect for unknown game reports

criezy:
In the ScummVM unknown game dialog there is some code to have button that opens a new ticket on our bug tracker with pre-filled information to report an unknown game. Currently this code is disabled as it directly accesses bugs.scummvm.org and this was deemed not very robust against future change to our bug tracker. You can see the discussion regarding this in https://github.com/scummvm/scummvm/pull/1160

There is also a comment in the code (https://github.com/scummvm/scummvm/blob/master/gui/unknown-game-dialog.cpp#L57) about it.

The idea here is to have the dialog generate a request in the form of `https://www.scummvm.org/unknowngame?engine=foo&description=bar` and our web site redirects this to our bug tracker with the proper query reformatting. If we make some changes to our bug tracker, we can at the same time update the redirect so that reporting unknown games in old ScummVM version will still properly work.

Note: To alley some fears, I should point out that reporting a game in this way will not automatically create a new ticket. This will open a form for a new ticket, with some field pre filled, but the user still has to validate it. Also the user will need to have an account and be logged in on bugs.scummvm.org.

I am really not familiar with web development in general and php in particular. So it is possible (and even likely) that this could be done in a cleaner, safer, and better way. So don't hesitate to comment on that.

If we decide to accept this PR, the next step will be to update the unknown game dialog to use this.

### DIFF
--- a/unknowngame.php
+++ b/unknowngame.php
@@ -1,0 +1,7 @@
+<?php
+    parse_str($_SERVER['QUERY_STRING'], $values);
+    # It is unlikely the engine name will be correct (e.g. missing capitalization) but
+    # try to use it nonetheless. The worst that can happen is that it will be ignored.
+    header("Location: https://bugs.scummvm.org/newticket?&type=enhancement&keywords=unknown-game&component=Engine:%20".$values['engine']."&description=".$values['description']);
+    exit();
+?>
--- a/unknowngame.php
+++ b/unknowngame.php
@@ -2,6 +2,6 @@
     parse_str($_SERVER['QUERY_STRING'], $values);
     # It is unlikely the engine name will be correct (e.g. missing capitalization) but
     # try to use it nonetheless. The worst that can happen is that it will be ignored.
-    header("Location: https://bugs.scummvm.org/newticket?&type=enhancement&keywords=unknown-game&component=Engine:%20".$values['engine']."&description=".$values['description']);
+    header("Location: https://bugs.scummvm.org/newticket?&type=enhancement&keywords=unknown-game&component=Engine:%20".urlencode($values['engine'])."&description=".urlencode($values['description']));
     exit();
 ?>
--- a/index.php
+++ b/index.php
@@ -102,6 +102,11 @@ $router = new \AltoRouter();
 // Custom match for Compatability ID.
 $router->addMatchTypes(array('cId' => '(DEV)|[0-9\.]++'));
 
+// Redirect for unknowgame
+$router->map('GET', '/unknowngame?[*:query]', function( $query ) {
+    require __DIR__ . '/unknowngame.php';
+});
+
 foreach ($pages as $key => $value) {
     $router->map('GET', $key, $value);
     $router->map('GET', $key . '/', $value);
@@ -109,8 +114,12 @@ foreach ($pages as $key => $value) {
 
 $match = $router->match();
 if ($match) {
+  if( is_array($match) && is_callable( $match['target'] ) ) {
+    call_user_func_array( $match['target'], $match['params'] );
+  } else {
     $page = new $match['target']();
     return $page->index($match['params']);
+  }
 } else {
   $page = new \ScummVM\Pages\NewsPage();
   return $page->index(array());


lotharsm:
Is this telepathy? I just started to dig out the updater dialog code again yesterday. Great :)

unknowngame.php
```diff
@@ -0,0 +1,7 @@
+<?php
+    parse_str($_SERVER['QUERY_STRING'], $values);
+    # It is unlikely the engine name will be correct (e.g. missing capitalization) but
+    # try to use it nonetheless. The worst that can happen is that it will be ignored.
+    header("Location: https://bugs.scummvm.org/newticket?&type=enhancement&keywords=unknown-game&component=Engine:%20".$values['engine']."&description=".$values['description']);
```
bluegr: %20 is the character for space... is it needed in the URL? Also, if the casing for the engine is wrong, you could convert it to lowercase here, right?

Mataniko:
This is probably not the right place to do this. Why not just open the URL directly from ScummVM? We haven't really made any changes to it in years.

We can probably apply redirection rules in the bug tracker vhost that will sort some of the problems discussed earlier, but you'd still need to be logged in.

unknowngame.php
criezy: The space is needed because the component names on our bug tracker are currently in the form `Engine: name`, and we need to match that if we want the component to be pre-selected in the new ticket form.

The comment on the capitalization is actually a bit outdated, but I forgot to update it after diving into it a bit deeper. The original query is generated by the code in ScummVM, and the engine name is what we get when calling the MetaEngine::getName() method. So for the component selection to work, we rely on this name matching the one used for the bug tracker component name. It actually works in many cases (for engine `Hugo` -> `Engine: Hugo` and `ADL` -> `Engine: ADL`). There are some cases where it doesn't though, but the issue is not necessarily capitalization (such as `Blade Runner` and `Engine: Bladerunner`). I think the idea to solve that is not to try to do smart things here, but rather to be consistent in the implementation of MetaEngine::getName() in ScummVM, and the naming of the component on the bug tracker. For example we might want to change `BladeRunnerMetaEngine::getName()` to return `Bladerunner` or to rename the component on the bug tracker to `Engine: Blade Runner` (in this specific case the later seems the best to me).

criezy:
@mataniko: The current implementation (that is commented out) in the unknown game dialog generates directly a URL for `http://bugs.scummvm.org/newticket?...`. If you read the discussion in https://github.com/scummvm/scummvm/pull/1160 you will see that Kirben raised the following issue
> If the bug report system formatting changes in the future, the report option gets broken.
And this is why the Report button in the dialog is currently not shown.

This PR is an attempt to mitigate the risk with bug tracker changes (such as when we migrated from sf.net bug tracker to our own track instance) and be more future proof, by disconnecting the URL generated by ScummVM from the actual bug tracker URL that needs to be called. There is less risk of `www.scummvm.org` disappearing than `bugs.scummvm.org` (for example if we decide to migrate to Github but tracker).

> We can probably apply redirection rules in the bug tracker vhost

Possibly. And this might be a better place.
But I suppose this assumes that we keep a bug tracker vhost just to do the redirection even if we decide in the future to move to a hosted bug tracker somewhere else?

I actually looked briefly at doing the redirection with a `.hataccess` rewrite rule before I noticed we no longer have a `.htaccess` file for the main website. And I think this would be feasible to do it that way.

> [...] but you'd still need to be logged in.

Which is what we want. We don't want anonymous report since we might need to request more information. And this also provides a filter to avoid spam reports.

Mataniko:
This will not be a problem for the foreseeable future.

I'm trying to remove much of the non-website logic from the website. Even if we migrate to a different bug tracker in the future we can keep the bugs.scummvm.org subdomain with the proper redirect until it stops getting hits and ScummVM is updated.

We have a separate endpoint for "admin" actions which this could also fit in, and since this can be handled as a rewriterule it would also fit better there

criezy:
> We have a separate endpoint for "admin" actions which this could also fit in

I have no idea what that means...

> And since this can be handled as a rewriterule it would also fit better there

Here is what I tried quickly yesterday when playing with `.htaccess`:
```
RewriteEngine On
RewriteCond %{REQUEST_URI} ^/unknowngame
RewriteCond %{QUERY_STRING} engine=([^&]*)&description=(.*)
RewriteRule (.*) https://bugs.scummvm.org/newticket?&type=enhancement&keywords=unknown-game&component=Engine:\ %1&description=%2 [R=302,L]
```

criezy:
I have done some further tests, and detected some issues.

1. With the unknown game.php approach for the redirect, if the description contained multiple lines (with end of line encoded as %0A) then we get an error: `Warning: Header may not contain more than a single header, new line detected in /Users/criezy/Dev/git-repos/scummvm-web/unknowngame.php on line 5`

This is fixed with the commit I have just pushed.

2. With the `.htaccess` approach, the rewrite rule encode the already encoded query element. So for example `description=a%20test` ends up being redirected as `description=a%2520test` which then gives a description in the new ticket form that is only partially decoded (we get `a%20test` instead of `a test`).

This can be fixed with a RewriteMap to decode the query strings. Those cannot be in the `.htaccess` and need to be in the server or host config file. We can for example use the following:
`RewriteMap decode int:unescape`

And then the `.htaccess` becomes:
```
RewriteEngine On
RewriteCond %{REQUEST_URI} ^/unknowngame
RewriteCond %{QUERY_STRING} engine=([^&]*)&description=(.*)
RewriteRule (.*) https://bugs.scummvm.org/newticket?&type=enhancement&keywords=unknown-game&component=Engine:\ ${decode:%1}&description=$(decode:%2} [R=302,L]
```

I have updated the `UnknownGameDialog` code in ScummVM to test this, and I have pushed the updated code so that others can also test (they will need to remove the `#if 0` in the code and change the URL to a test server with a redirection in place). In my tests both approaches seem to work now to redirect the request to our bug tracker.

Mataniko:
The site should not use `.htaccess` files. 
If you want to do a redirect, please do it in the Apache config file for scummvm.org redirects.

bluegr:
> The site should not use `.htaccess` files.
> If you want to do a redirect, please do it in the Apache config file for scummvm.org redirects.

@Mataniko:
1) Why should we not use .htaccess files at all?
2) Why is the Apache config better for redirects?
3) What do you mean with “admin” actions?

Mataniko:
1. .htaccess overrides server admin directives, making it a pain to manage the site.
2. https://httpd.apache.org/docs/2.4/howto/htaccess.html#when
3. Please look at the apache config, specifically scummvm-redirect.conf and scummvm-admin.conf

criezy:
@Mataniko Thank you for the link and pointers. That makes sense.
The content of scummvm-admin.conf went over my head. But it seems I can mostly ignore it anyway? Or is it relevant in some way, other than to illustrate you first point?

I see we have two conf-fragments with rewrite rules: `scummvm-rewrite.conf` and `scummvm-mirrors.conf`. What is the difference between the two? And to add rewrite rules for bugs.scummvm.org, should I add them to one of those two files, or is it better to create a new `bugs-rewrite.conf` fragment?

Mataniko:
`scummvm-rewrite` is for scummvm.org rewrite rules, `scummvm-mirrors.conf` is for mirrorbrain.

You can put the same rewrite rule that you'd put in `.htaccess` in `scummvm-rewrite`. There's no need to put a rewrite rule for bugs.scummvm.org because there you can open the URL directly.

bluegr:
Many thanks for your answers, @Mataniko :) they cleared things up

criezy:
> There's no need to put a rewrite rule for bugs.scummvm.org

Oh OK. I though you argued above that the unknown game dialog should use `https://bugs.scummvm.org/unknowngame?engine=foo&description=bar` instead of `https://www.scummvm.org/unknowngame?engine=foo&description=bar`. So that's where my question was coming from.

criezy:
As `scummvm-setup` cannot be forked, no pull request can be created, so I will have to push directly to master. But before I do let's confirm here that the change I plans is good.

Reading the apache doc I saw that we can use the NE flag to stop percent sign being escaped, which removes the need to decode the query strings for the rewrite rule. So there is actually no need to add a RewriteMap. That simplifies the change as it simply means adding the following to the end of the  `scummvm-rewrite.conf`:
```
# Unknown game reports
RewriteCond %{REQUEST_URI} ^/unknowngame$
RewriteCond %{QUERY_STRING} engine=([^&]*)&description=(.*)
RewriteRule ^(.*)$ https://bugs.scummvm.org/newticket?type=enhancement&keywords=unknown-game&component=Engine:\ %1&description=%2 [L,NE]
```

In particular I am not sure if a `R` flag should be used. I dropped it after noticing that in the current `scummvm-rewrite.conf` there is no `R` for most of the rewrite rules, and checking that it indeed works without one. But I don't really know what the impact is.

Mataniko:
You can simplify it:
`RewriteRule ^/unknowngame\?engine=(.+)&description=(.+)$ https://bugs.scummvm.org/newticket?type=enhancement&keywords=unknown-game&component=Engine: $1&description=$2 [L,NE]`

criezy:
Hmmm. That doesn't work on my test server.

It gives me a `Internal Server Error`, but this one can easily be fixed by escaping the space for the component name, or using `%20` instead of a space and escaping the % (so either `component=Engine:\ $1` or `component=Engine:\%20$1`).

But then that still doesn't work as the rule for some reason is not matched and I get an error `The requested URL /unknowngame was not found on this server`. I tried various variations, but I can't find anything that works, and I have no idea what is wrong.

criezy:
Oh, never mind. We just need to drop the leading forward slash. So this one works:
`RewriteRule ^unknowngame\?engine=(.*)&description=(.+)$ https://bugs.scummvm.org/newticket?type=enhancement&keywords=unknown-game&component=Engine:\%20$1&description=$2 [L,NE]`

Mataniko:
sorry, you're correct. the forward slash is not necessary, cut and paste error :)

Closing this PR.

PR #102, opened 2019-09-27
WEB: Promote Blade Runner compatiblity to excellent

lotharsm:
As discussed a few weeks ago, this PR promotes Blade Runner to "Excellent" on our compatibility list.

Submitting as PR in case someone objects - feedback welcome.

### DIFF
--- a/data/compatibility/compat-DEV.xml
+++ b/data/compatibility/compat-DEV.xml
@@ -1092,9 +1092,9 @@
 			<game>
 				<name>Blade Runner</name>
 				<target>bladerunner</target>
-				<support_level>good</support_level>
+				<support_level>excellent</support_level>
 				<notes>
-					Game is completable.
+					No known issues, game is completable.
 				</notes>
 			</game>
 			<game>


dafioram:
I vote good for 2.1.0, it will get a lot more tested when people play it because of the release.

sev-:
Historically 'excellent' was reserved to the games we had access to the original sources for. But now I see that bunch of SCI games were put to that category which is kind of misleading.

I am not sure, though, whether we should follow that categorisation. Thoughts?

peterkohaut:
Since the announcement we didn't have many issues related to the engine/game it self (at least the without cut content).  Most of them were with ports android, ps3, amiga. And antoniou79 was able to fix many issues which are present in the original game.

To be honest, this ranking is a bit misleading as there is no formal definition and I don't think access to the source code is a guarantee for "excellent" rank as we all make coding errors or various ports might be unhappy about something.

But to avoid confusion we can keep it "good" as it can be changed any time :)

eriktorbjorn:
On 27/09/2019 23.23, Eugene Sandulenko wrote:
> Historically 'excellent' was reserved to the games we had access to the 
> original sources for. But now I see that bunch of SCI games were put to 
> that category which is kind of misleading.
> 
> I am not sure, though, whether we should follow that categorisation. 
> Thoughts?

I thought that standard was broken long ago, though. Or did we receive 
the source code for Toonstruck? It's listed as "Excellent", but last 
time I checked it had graphical glitches. (The ones I saw when I played 
through the early parts a couple of years ago were #6867 and #6869. The 
first one of them can be easily reproduced since it happens in the very 
first room of the game.)

Torbjörn Andersson

Thunderforge:
As an end user who is more interested in helping with documentation than source code, it has never been obvious at all to me that "Excellent" has ever meant "we have the source code". I had assumed it meant "there are minor, if any, bugs remaining in the game compared to the original engine."

The labels "Untested", "Broken", "Bugged", "Good", and "Excellent" all seem to me like indicators of quality and stability. If the game is working perfectly compared to the original, then I don't care behind the scenes whether or not you have access to the source code.

sev-:
Well, you are not taking into account the history of the project. Just take a look at how it was attributed several years ago: https://www.scummvm.org/compatibility/1.6.0/

And honestly, we are doing the project for fun and not to please somebody's wishes or desires. I am referring to the part where you diminishing our work by expressing that you "don't care [about the scummvm developers' affairs]".

criezy:
I think what he meant is that from a user perspective the main criteria is how well the game plays. Whether we had access to the source code, or reimplemented the engine using reverse engineering is not really something that users would look at. I don't feel that there was any intent to diminish the work that goes into adding support for a game.

And I would tend to agree with this. When I see the compatibility page on the website, I would assume that the rating is an indicator of how well the game is runs, and if there are any major or annoying issues remaining. So personally I think we should not limit the Excellent rating to games for which we have the original source code. If support for a reverse-engineered game is very good with no know issues, I don't see why it could not be rated as excellent.

Thunderforge:
Yes, I was describing my own perspective as a user perspective, not as someone who has any history of the project. I apologize that my words were misunderstood to come across as saying that I "don't care [about the scummvm developers' affairs]". Quite the contrary! I want to make sure that the levels indicated on the version pages convey the quality work that developers have produced. As-is, I feel that the labels are misleading and lead to users such as me misinterpreting them into thinking that less progress has been done than in actuality.

sev-:
In fact, I support the idea. I am more proponent of something like:

"Good", "Minor issues", "Major issues". And on top of that stays our requirement for the game to be completable in order to enter the lists at all.

lotharsm:
@sev- Sounds like a very good proposal!

Thunderforge:
> "Good", "Minor issues", "Major issues". And on top of that stays our requirement for the game to be completable in order to enter the lists at all.
> 
> PS. And "Excellent" when we improve the game or fix the number of original bugs.

@sev I like the descriptors of “Major Issues” and “Minor Issues”, as it is very clear what it means. But what would “Good” mean in this context? Is it worse off than “Major Issues”? Or better off than “Minor Issues”? Perhaps a more descriptive phrase could be used. If it’s worse, then maybe “Completable” or “Severe Issues”. If it’s better, then maybe “Trivial Issues”.

Also, while it’s clear that “Excellent” means that there is really good compatibility with the game, I’m not sure that the difference between “Minor Issues” and “Excellent” is very clear. I wonder if there is a word that could be used to express “Issues in original fixed” or “Better than original”.

Overall though, I like these labels better, and agree with the desire to only list completable games. It might be good to say something like “these games are completable. Click here for a list or incomplete games.”

SupSuper:
Here's my suggestions:

- Broken: game is detected but impossible to complete
- Bugged: game is completable but has major features broken / missing (i.e. an inexperienced user would have to know what to avoid or they might get stuck)
- Good: game is completable and only minor features broken / missing (i.e. an inexperienced user won't run into anything game-breaking)
- Excellent: game has every feature present and is properly integrated into ScummVM

sev-:
The broken games should be either fixed or removed from the list. There is no sense to keep them there.

Calling a game with broken features as Good is questionable. Normally, we add games to the list once the devs have all known features implemented.

This is not a GAME LIST, but a ScummVM Compatibility list. We cannot add all possible games here, or half-baked ones. This discussion deviated significantly.

Mataniko:
Closing this PR since it stagnated.

PR #103, opened 2019-10-10
Adding unsupported Backyard Sports games

Thunderforge:
Consensus in Discord seems to be that [these SCUMM games](https://wiki.scummvm.org/index.php?title=Humongous_Entertainment/Games#Junior_Sports_games) are unsupported because of missing functionality. Adding them to the list for completeness.

### DIFF
--- a/data/compatibility/compat-DEV.xml
+++ b/data/compatibility/compat-DEV.xml
@@ -1615,6 +1615,14 @@
 				</notes>
 			</game>
 			<game>
+				<name>Backyard Basketball</name>
+				<target>basketball</target>
+				<support_level>broken</support_level>
+				<notes>
+					Able to be run, but large portions of gameplay are missing.
+				</notes>
+			</game>
+			<game>
 				<name>Backyard Football</name>
 				<target>football</target>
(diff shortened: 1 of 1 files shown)


sev-:
Thanks for your contribution, but those games are not "broken", they are not in our compatibility list. We have dozens of the games still in making, but we add them to the compatibility page only once they're ready at least for the public testing.

The rule of the thumb is: no game in the compatibility list, no run.

Thunderforge:
@sev- The compatibility page has "Blue's Treasure Hunt" (another Humongous Entertainment SCUMM game) listed as "Bugged". What is the difference between something like that and this game?

More generally, under what cases would a game be on the list but marked as "Bugged" (and can we add that to the documentation described in #104)?

sev-:
Blue's Treasure Hunt has some visual issues, but is fully playable. The Backyard Sports games you refer to are simply incomplete in terms of their code.

If you have time and energy, we have access to the source code of the missing pieces, so it is possible to complete those. Most of that code is hairy floating-point calculations for the ball trajectories and AI for the players.

PR #107, opened 2019-10-20
WEB: Add compatibility for incomplete Backyard Sports games

SupSuper:
Currently these games are detected by ScummVM even though the core gameplay is missing. This confuses some users, since other Backyard games are listed and fully playable, so they assume they all must be.

I suggest explicitly marking them as "broken". Alternatively, the detection tables could be changed to hide them from users or flag them as "unstable".

### DIFF
--- a/data/compatibility/compat-DEV.xml
+++ b/data/compatibility/compat-DEV.xml
@@ -1808,6 +1808,15 @@
 				</notes>
 			</game>
 			<game>
+				<name>Backyard Basketball</name>
+				<target>basketball</target>
+				<support_level>broken</support_level>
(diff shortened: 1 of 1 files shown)


tsoliman:
It is indeed confusing, hence #103

SupSuper:
Ah I missed that. @sev Maybe something could be done about the detection tables then? Since games "in the making" are normally flagged as UNSTABLE or excluded from stable builds.

sev-:
No, we have ONLY supported games in the compatibility list. If we start adding all detected but unsupported games, it will bloat the list several times.

PR #113, opened 2020-03-23
DATA: Changing encoding of compatibility XMLs to UTF-8

### LINKED ISSUE #112: Touché rendered as TouchÃ© on Compatibility page
The game *Touché: The Adventures of the Fifth Musketeer* appears on the ScummVM [compatibility page](https://www.scummvm.org/compatibility/) as

image

> TouchÃ©: The Adventures of the Fifth Musketeer

In other words, the `é` symbol is being improperly displayed as `Ã©`. In [data/compatibility/compat-DEV.xml](https://github.com/scummvm/scummvm-web/blob/045d25ed835ac79bade53d9d23bd6aff48899827/data/compatibility/compat-DEV.xml#L1725), it's listed correctly, so the issue must be with rendering.

Tested on macOS Safari and Firefox.

Thunderforge:
Fixes #112

image

### DIFF
--- a/data/compatibility/compat-2.1.1.xml
+++ b/data/compatibility/compat-2.1.1.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="ISO-8859-1"?>
+<?xml version="1.0" encoding="UTF-8"?>
 <compatibility xmlns:h="http://www.w3.org/TR/xhtml1/">
 	<company>
 		<name>LucasArts</name>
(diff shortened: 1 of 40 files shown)


data/compatibility/compat-2.1.1.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="ISO-8859-1"?>
+<?xml version="1.0" encoding="UTF-8"?>
```
Thunderforge: All other `compat-*.xml` files still have ISO-8859-1 as their encoding, but only 2.1.1 and DEV have "Touché", rather than "Touche" as their game name.

lotharsm: Even though it's not necessary to fix an actual encoding/rendering issue, I recommend changing the encoding of the remaining files too, so we have a consistent acoding across all files.

Thunderforge: @lotharsm Done.

lotharsm:
Thanks, merging!

PR #115, opened 2020-03-26
DATA: Fixing/removing broken links

Thunderforge:
* Removing dead links to official and unofficial ports
  * All are dead except the GP32 port, which hasn't been updated in 6 years
* Changing hqx link to Wikipedia, since the main site is dead
* Correcting link to Ogg Vorbis

### DIFF
--- a/data/links.xml
+++ b/data/links.xml
@@ -1,34 +1,6 @@
 <?xml version="1.0" encoding="ISO-8859-1"?>
 <external_links xmlns:h="http://www.w3.org/TR/html4/">
 	<group>
-		<name>Official ports</name>
-		<description>Some additional links for ports that have been merged into our repository</description>
-		<link>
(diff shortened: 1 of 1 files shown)


data/links.xml
```diff
@@ -58,7 +30,7 @@
 		</link>
 		<link>
 			<name>hq2x/3x/4x</name>
-			<url>http://www.hiend3d.com/hq3x.html</url>
+			<url>https://en.wikipedia.org/wiki/Hqx</url>
```
sev-: It is better to change it to Web Archive: https://web.archive.org/web/20131114143602/http://www.hiend3d.com/hq3x.html

Thunderforge: @sev- Done.

sev-:
Thanks!

Issue #110, opened 2020-02-25
Combine game compatibility list into one table with sortable columns


Thunderforge:
## Problem

As the list of games on the [compatibility page](https://www.scummvm.org/compatibility/) grows longer and longer, it becomes increasingly difficult to find games that one might be interested in. This is made all the more difficult because each table does not allow sorting by column. A the moment we have:

- Number of games: 257
- Number of tables: 12
- Number of games in “Other”: 65

Say you want to find out if *Once Upon A Time: Little Red Riding Hood* is supported. Is it under "Animation Magic", "Coktel Vision", "Living Books", or “Other”? With our current system, you have to check up to four tables.

## Initial suggestion for a solution

I’d like to propose that we reorganize the compatibility page. My initial thought is to have just one table with the following columns:

- Game Full Name
- Game Short Name
- Company
- Engine
- Data Files
- Support Level

Then have each column sortable in ascending or descending order alphabetically.

## Advantages:

- It is still easy to find all games from the big publishers (e.g. LucasArts, Sierra)
- Multiple games with dissimilar names from small companies (e.g. *Myst* and *Riven* by Cyan) are grouped together when sorting by companies
- Alphabetical sorting by name makes it easier to find a game if you know the name, but not the company
- Sorting by engine means it’s easier to see all SCUMM games, all Composer games, etc, giving a general state of the engine progress.

## Disadvantages:

- Series containing games that were released by multiple companies (e.g. Living Books, AdventureSoft/HorrorSoft) are no longer grouped together
- Adding more columns may make the page more difficult to read on mobile devices

Mataniko:
The biggest hurdle here is that we're missing a lot of that metadata, and that's a completely manual process to do.

Generally the current XML needs to be scrapped into something a little more structured (YAML files describing games, engines, compatibility, etc.). Only once that's done, we can work on the front end to include sorting, grouping, searching, whatever (ScummVM will soon support a magnitude more of games when some of the in-progress engines are finished)

If you'd like to start tackling the metadata problem that would be greatly appreciated.

Thunderforge:
@Mataniko I'm tentatively interested in doing the work to change the metadata to YAML. I have a few questions first:

1. Will this need to be changed for the most recent version only or for all versions?
1. What is the best way to contribute these? As a PR?
1. What would be the best way to discuss how these YAML files will be structured? This Issue or something else?

Mataniko:
@Thunderforge 

[...]

PR #116, opened 2020-03-26
DATA: Converting Game Demos page data from XML to YAML

Thunderforge:
Since it appears that #110 will require compatibility data to be changed from XML to YAML, I wanted to practice and change a simpler page from XML to YAML. If merged, I'm willing to convert other pages as well (e.g. Links, Press Articles, Screenshots, FAQs).


Thunderforge:
@lotharsm Thank you! I hope to convert more pages soon, as well as port my changes to [residualvm-web](https://github.com/residualvm/residualvm-web).

PR #117, opened 2020-03-31
WEB: Add link to browse frs in Old Versions section of the Download page

criezy:
In the past few months there have been several users trying to find old releases of ScummVM and sometimes resorting to getting them from archive.org (see for example [this forum post](https://forums.scummvm.org/viewtopic.php?p=88339#p88339)).

So I was thinking that maybe it would be a good idea to add a link to https://www.scummvm.org/frs/scummvm/ in the Older Versions sections of our Download page. What do you think? Also do you see a better way to do it than splitting the string in three segments? I took the idea from what is already done for the release note link Release binaries section of the same page, but that's not ideal for translators...

Here is what it looks like:
image


sev-:
I love it!

PR #118, opened 2020-04-02
DATA: Changing all XML files to UTF-8

Thunderforge:
The compatibility files have already been changed through #113. Changing the rest for consistency. Compared the pages and didn't see any differences.


sev-:
Makes sense.

PR #126, opened 2020-04-04
(WIP) DATA: Converting compatibility data to single YAML file

### LINKED ISSUE #121: Proposal: Reorganize compatibility data into one YAML file
This is a proposal to reorganize the [compatibility data](https://github.com/scummvm/scummvm-web/tree/master/data/compatibility) on the ScummVM website. Feedback is encouraged.

## Problem
At the time of this writing, there are 40 different files detailing the compatibility of various versions of ScummVM. While this does make it easy to see what was supported in a particular version, there are some major downsides:

- Dozens of files makes updating difficult and allows for inconsistencies
   - Changing the XML structure can require changing up to 40 different files
   - If new metadata tags are added without changing the structure, then old compatibility pages have empty columns (e.g. [1.6.0 compatibility](https://www.scummvm.org/compatibility/1.6.0/) has an empty "Data Files" column)
   - Game names are inconsistent between pages (e.g. [2.1.1](https://www.scummvm.org/compatibility/2.1.1/) says "Indiana Jones and the Fate of Atlantis", but [0.4.0](https://www.scummvm.org/compatibility/0.4.0/) says "Indiana Jones 4 and the Fate of Atlantis")
   - Tags are inconsistent between pages (e.g. some have `<percent>`, others have `<support_level>`)
- The current `<support_level>` tag is not working well
  - The meaning of each level has been ambiguous/contentious (#102) and undocumented (#104)
  - The `Untested`, `Broken`, and `Bugged` levels are largely unused, and developers have [indicated](https://github.com/scummvm/scummvm-web/pull/102#issuecomment-544666480) that they don't want to continue to use them (and PRs to add them like #107 have been rejected)
  - Having two levels of distinction (`Good` and `Excellent`) means it's more difficult to contrast different levels of support
- XML is not a preferred data format, with YAML being a preferred alternative
  - The XML format is not suitable for sorting or categorization (#110)
  - Data for other pages on the website have been converted to the YAML format (e.g. Game Demos via #116), and this is a trend we seem to want to continue

### A new `since` metadata value
A new `since` metadata value would be included with each game entry (matching the [value on the Wiki](https://wiki.scummvm.org/index.php?title=Maniac_Mansion)) that would indicate which version of ScummVM added support. For instance, Maniac Mansion would have `<since>0.5.0</since>` (YAML: `since: "0.5.0"`).

This would *replace* both the `<percentage>` tag and the `<support_level>` tag. While this does mean that data is lost (see "Drawbacks" section), it also means that there are no more debates on what to classify each version. Anybody who wants to know if there are lots of issues or not can simply look at the bug tracker.

I checked a few games that are listed as `Bugged` or `Broken`, and the Wiki says they are unsupported, and therefore won't have a value for `since`. I suggest they all be removed.

### One file for all games ever supported
The `since` value allows us to have *one* file for compatibility containing all games. It would contain all games that have been supported in any version of ScummVM. Then the list can be filtered (via PHP or in the frontend) to whichever ones are supported in the requested version.

For instance, if the user wants to see which ones are supported in ScummVM version 0.5.0, the full list would be filtered to any games that have a `since` value of 0.5.0 or earlier. The `CompatabilityModel.php` file already includes a `compareVersions()` plugin to compare semantic versions, so I don't think there is any additional work needed.

### YAML file format

The above could be done in XML, but this would be a good time to switch over to YAML. This would also give us the opportunity to add a few fields such as `engine` or `versionsSupported`. For instance:

```
- game:
    name: Maniac Mansion
    shortName: maniac
    company: LucasArts
    engine: SCUMM
    datafiles: Maniac_Mansion_.28Original_or_Enhanced.29
    since: 0.5.0
    versionsSupported:
        - Amiga
          # If necessary, we can add since information to each version
           - since: x.y.z
        - Apple II
        - Atari ST
        - Commodore 64
        - Macintosh
        - NES
        - PC
```

## Advantages

 - With one file, keeping game information consistent between releases is easy
 - New metadata like `engine` can be added easily since there is only one file
 - Game names will be consistent between versions
 - There are no more debates on what to set the subjective `percentage` or `support_level` to, since it's replaced by the objective `since` that matches the Wiki
 - The YAML file format will allow for new website features like sortable tables and filtering (#110)

## Disadvantages

 - Historical data for `percentage` or `support_level` is lost (e.g. the website no longer shows that version 1.6.0 had 90% compatibility with Maniac Mansion)
   - I'm guessing that this isn't a big loss, but I don't know for sure. Perhaps site analytics or a straw poll in IRC/Discord would be helpful?
 - A fair amount of work is needed to make this conversion (I'm willing to volunteer)

Thunderforge:
Fixes #121.

This is my work in progress branch for converting compatibility data to a single YAML file, per the proposal of #121. I will be iteratively making changes to it. The general process is as follows:

1. Start with the compat-DEV.xml, since it has the latest information on supported games
1. Convert to YAML
1. Reorganize per the proposal
1. Add `since` tags
1. Remove unnecessary notes and other obsolete data
1. Make any other changes deemed necessary
1. Link up to the compatibility page on the website

At this point, the website will largely look the same, but the data will be easier to manage. Future PRs will enhance the compatibility page (e.g. adding the ability to filter results).

### DIFF
--- a/data/compatibility.yaml
+++ b/data/compatibility.yaml
@@ -236,7 +236,7 @@ companies:
             - name: "DOS"
           unsupported:
             - name: "Commodore 64"
-              notes: "Doesn't use AGOS, so will never be supported"
+              notes: "Doesn't use AGOS"
 
       - name: "Elvira II - The Jaws of Cerberus"
         target: "elvira2"
@@ -251,7 +251,7 @@ companies:
               notes: "No sound effects"
           unsupported:
             - name: "Commodore 64"
-              notes: "Doesn't use AGOS, so will never be supported"
+              notes: "Doesn't use AGOS"
 
       - name: "Personal Nightmare"
         target: "pn"
@@ -1712,7 +1712,7 @@ companies:
             - name: "Windows"
           unsupported:
             - name: "CD-i"
-              notes: "Doesn't use Groovie, so it will never be supported."
+              notes: "Doesn't use Groovie"
 
       - name: "TeenAgent"
         target: "teenagent"
@@ -2036,7 +2036,7 @@ companies:
             - name: "Windows"
           unsupported:
             - name: "PlayStation"
-              notes: "Doesn't use SCUMM, so will never be supported."
+              notes: "Doesn't use SCUMM"
 
       - name: "Pajama Sam's Lost & Found"
         target: "lost"
--- a/data/compatibility.yaml
+++ b/data/compatibility.yaml
@@ -1,0 +1,2392 @@
+companies:
+  - name: "LucasArts"
+    games:
+      - name: "Maniac Mansion"
+        target: "maniac"
+        company: "LucasArts"
+        engine: "SCUMM"
+        since: "0.5.0"
+        datafiles: "Maniac_Mansion_.28Original_or_Enhanced.29"
+        platforms:
+          supported:
+            - name: "Amiga"
+            - name: "Apple II"
+            - name: "Atari ST"
+            - name: "Commodore 64"
+            - name: "Macintosh"
+            - name: "NES"
+              notes: "Minor graphical glitches"
+            - name: "PC"
+
+      - name: "Zak McKracken and the Alien Mindbenders"
+        target: "zak"
+        company: "LucasArts"
+        engine: "SCUMM"
+        support_level: "good"
+        datafiles: "Zak_McKracken_and_the_Alien_Mindbenders"
+        platforms:
+          supported:
+            - name: "Amiga"
+            - name: "Atari ST"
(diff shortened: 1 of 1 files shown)


lotharsm:
@Mataniko can you take a look at this? I still like the idea to convert everything into modern YAML files.

Mataniko:
I don't think we should do the conversion as is without fixing the underlying issues we have with the data structure. I've had extensive conversations on this with @Thunderforge over Discord. I'll loop you in.

Mataniko:
Obsolete with the new data model

PR #130, opened 2020-08-31
WEB: Refactor L10N handling

Mataniko:
This PR updates the L10N handling to use full locale codes and automatically determine what languages are available in the sidebar and the locale date formatting.


Thunderforge:
There's a lot of files in here, but I think I gt the gist of it and I don't see any issues. I'm okay with merging once merge conflicts are resolved.

sev- (approved):
Looks good to me. However, we were using term 'internationalization', not 'location', but that doesn't matter (I am referring to the commit log descriptions)l

Mataniko:
I changed the term from Internationalization to Localization which is more appropriate to what we're doing. Moving forward commits would be L10N.

files were moved to make loading more consistent between the default (en_US) locale and the rest.

PR #131, opened 2020-09-12
WEB: Modernize the data model & add caching

Mataniko:
This PR updates the current data model and moves most data into a database-like structure.

For example, versions are no longer read from filenames, various names for games, companies and engines now pull from their respective data files and cross referenced.

The first page that this can be seen is in the compatibility page, where we now have a much simpler format for compatibility data (see compatibility.yaml) where we only need to enter the deltas between versions, and not copy the entire file again. By having more structured data, we now auto generate most of the game notes - leaving devs to only add pertinent information.

Once other pages are updated (demos, screenshots) we will also have consistency in naming across pages.

There's more work to do before merging this: safety checks, missing data validation, bug squashing, naming and general data correction - but as of now the site is runnable with the new model. 

I encourage you to try it out, and suggest improvements and corrections.


Mataniko:
With the latest commit, I think this should be good to merge.

Mataniko:
Taking back what I said - I added a script to pull data directly from the spreadsheet and into the site - so updates will not require manually updating yaml files. Instead all work can be done in sheets, as a visual database.

data/platforms.yaml
```diff
@@ -0,0 +1,96 @@
...
+  name: FM Towns
+  enum: kPlatformFMTowns
+- id: ios
+  name: iOS
+  enum: kPlatformIOS
+- id: linux
+  name: Linux
+  enum: ""
+- id: mac
+  name: Macintosh
```
Thunderforge, line 46: I suggest "Classic Mac OS" because that's [the name used by Wikipedia](https://en.wikipedia.org/wiki/Classic_Mac_OS). Similarly, the id would be `classic-macos`. This would avoid any potential confusion with "Modern" macOS, both of which are sometimes informally called "Macintosh".

data/series.yaml
```diff
@@ -0,0 +1,85 @@
+# This is a generated file, please do not edit manually
+---
+- id: dw
+  name: Discworld
+- id: indy
+  name: Indian Jones
```
Thunderforge, line 6: Should be Indiana Jones.

data/companies.yaml
```diff
@@ -0,0 +1,744 @@
+# This is a generated file, please do not edit manually
+---
+- id: 1c
+  name: 1C Company
+  alt_name: ""
```
Thunderforge, line 5: I assume that the `""` is necessary if it's blank? I would prefer omitting it if it's optional, since quotes aren't used elsewhere.

data/companies.yaml
```diff
@@ -0,0 +1,744 @@
...
+- id: virgin
+  name: Virgin Interactive
+  alt_name: ""
+- id: westwood
+  name: Westwood Studios
+  alt_name: ""
+- id: zojoi
+  name: Zojoi
+  alt_name: ""
+- id: ""
```
Thunderforge, line 456: Why do these have no id? Did we just not get around to giving one?

data/compatibility.yaml
```diff
@@ -0,0 +1,1791 @@
+# This is a generated file, please do not edit manually
+---
+- id: maniac
+  support: good
+  notes: '- Minor graphical glitches in NES version'
```
Thunderforge, line 5: I think we can get rid of the leading `- `.

data/compatibility.yaml
```diff
@@ -0,0 +1,1791 @@
...
+  version: 2.0.0
+  stable_platforms: amiga,apple2,atarist,c64,dos,nes
+  unstable_platforms: ""
```
Thunderforge, line 8: I like that we have stable and unstable platforms separated.

data/compatibility.yaml
```diff
@@ -0,0 +1,1791 @@
+# This is a generated file, please do not edit manually
+---
+- id: maniac
+  support: good
+  notes: '- Minor graphical glitches in NES version'
+  version: 2.0.0
```
Thunderforge, line 6: What does `version` mean here? The version of the game? The version of ScummVM it was first supported in?

Assuming it's the first version that ScummVM supported it, I think it would be better to have this property called `since_version` instead.

data/compatibility.yaml
```diff
@@ -0,0 +1,1791 @@
...
+  unstable_platforms: ""
+- id: zak
+  support: good
+  notes: '- FM-TOWNS Kanji version requires the FM-TOWNS Font ROM'
+  version: 2.0.0
+  stable_platforms: amiga,atarist,c64,dos,fmtowns
+  unstable_platforms: ""
+- id: indy3
+  support: good
+  notes: '- There is no support for the Macintosh interface - FM-TOWNS Kanji version
```
Thunderforge, line 17: I don't know that this "bullet points" works well in a string. It would be good to have some other structure. Maybe a YAML list?

```yaml
notes:
    - There is no support for the Macintosh interface
    - FM-TOWNS Kanji version
```
This would also avoid the issue of a leading `–`.

data/compatibility.yaml
```diff
@@ -0,0 +1,1791 @@
+# This is a generated file, please do not edit manually
+---
+- id: maniac
+  support: good
```
Thunderforge, line 4: I'm still of the opinion that "support" is a useless metric since they all are either "Good" or "Excellent" and we can't seem to agree on what the difference is between them, but I'm willing to fight that battle another day and not let it hold up this review.

data/compatibility.yaml
```diff
@@ -0,0 +1,1791 @@
...
+  version: 2.0.0
+  stable_platforms: amiga,atarist,dos,dos,fmtowns,mac,segacd
+  unstable_platforms: ""
+- id: monkey2
+  support: excellent
+  notes: '- Demo version often crashes due to missing resources, since it was never
+    meant to be playable - No support for playing back the recorded file of gameplay
+    in demo version - FM-TOWNS Kanji version requires the FM-TOWNS Font ROM - Loading
+    of original instrument samples for Amiga version not implemented, so music differs
+    from original interpreter.\'
```
Thunderforge, line 51: Trailing `\` should be removed.

public_html/index.php
```diff
@@ -1,6 +1,16 @@
 <?php
 namespace ScummVM;
 
+/**
+ * Development only - when using PHP built in web server
```
Thunderforge, line 5: And what does this do that we only want it to happen when run in development? A comment explaining the functionality would be helpful.

include/Pages/CompatibilityPage.php
```diff
@@ -87,14 +122,21 @@ public function getGame($target, $version, $oldLayout)
...
-        if ($version != 'DEV') {
+        if ($version !== 'DEV') {
             $key = array_search($version, $versions);
             unset($versions[$key]);
         }
-        $filename = DIR_COMPAT . "/compat-{$version}.xml";
-        $last_updated = date("F d, Y", @filemtime($filename));
-        $compat_data = CompatibilityModel::getAllData($version);
 
+        $compareVersion = $version === 'DEV' ? '9.9.9' : $version;
```
Thunderforge, line 130: If we ever changed our versioning scheme (such to Chrome-style where we increase the full version each release and get up to like version 75) this will break. Or if we just keep this code long enough that we do see ScummVM 9.10.0. I would prefer a solution that is more robust for that, but I won't hold up the PR for it.

include/Objects/LegacyCompatGame.php
```diff
@@ -2,10 +2,10 @@
 namespace ScummVM\Objects;
 
 /**
- * The CompatGame class represents a game on the compatibility charts on the
+ * The LegacyCompatGame class represents a game on the compatibility charts on the
```
Thunderforge, line 5: I assume that `Legacy` in the name refers to the class being legacy in contrast to a new `CompatGame` class? It would be good to add the new replacement in the comments.

include/Objects/Engine.php
```diff
@@ -0,0 +1,40 @@
+<?php
+namespace ScummVM\Objects;
+
+/**
+ * The Engine object represents a game engine supported by ScummVM
```
Thunderforge, line 5: Are they all supported? I thought there were a few WIP engines included. If they are indeed only supported engines, then ignore this comment.

include/Objects/Game.php
```diff
@@ -0,0 +1,61 @@
+<?php
+namespace ScummVM\Objects;
+
+/**
+ * The Game object represents a game known to ScummVM
```
Thunderforge, line 5: This is analogous to the Engine class, but that one says it's supported engines while this just says it's "known to", meaning they can be supported or unsupported. It would be good to clarify (and be consistent if possible).

include/Objects/Compatibility.php
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        if ($this->unstablePlatforms) {
+            $notes .= "**Unsupported Platforms:**\n";
+            $notes .= "$this->unstablePlatforms\n";
+        }
+
+        if ($this->notes) {
+            $notes .= "**Additional Notes:**\n";
+            $notes .= str_replace("- ", "\n- ", $this->notes) . "\n";
```
Thunderforge, line 79: This is temporary, right? It seems that it would be better to long-term format the data so that such conversions are unnecessary.

include/Objects/Compatibility.php
```diff
@@ -0,0 +1,121 @@
+<?php
+namespace ScummVM\Objects;
+
+use Erusev\Parsedown;
+use ScummVM\Objects\LegacyCompatGame;
+
+/**
+ * The Compatibilith object represents a game on the compatibility charts on the
```
Thunderforge, line 8: Typo: `Compatibilith` → `Compatibility`.

include/Objects/Company.php
```diff
@@ -0,0 +1,32 @@
+<?php
+namespace ScummVM\Objects;
+
+/**
+ * The Company object represents a companies that created games supported
```
Thunderforge, line 5: `represents a companies` → `represents the companies`.

include/Models/VersionsModel.php
```diff
@@ -0,0 +1,30 @@
...
+        $data = [];
+        foreach ($versions as $version) {
+            $obj = new Version($version);
+            $data[$obj->getId()] = $obj;
+        }
+
+        \uasort($data, "version_compare");
+
+        if (!array_key_exists('DEV', $data)) {
+            $data['DEV'] = new Version(["id" => 'DEV', "date" => "1/1/2099"]);
```
Thunderforge, line 26: I assume this is necessary to keep other things from breaking; I'm in favor of instead handling undefined data specially instead of creating arbitrarily large values, but I won't hold the PR up over it.

include/Models/LegacyCompatibilityModel.php
```diff
@@ -0,0 +1,90 @@
...
+
+use ScummVM\Objects\LegacyCompatGame;
+use ScummVM\XMLParser;
+
+/**
+ * The CompatibilityModel class will generate LegacyCompatGame objects.
+ */
+abstract class LegacyCompatibilityModel extends BasicModel
+{
+    const NO_VERSION = 'No version specified.';
```
Thunderforge, line 12: See, this seems like a better way to handle things than versions of `9.9.9`.

include/Models/CompaniesModel.php
```diff
@@ -0,0 +1,23 @@
+<?php
+namespace ScummVM\Models;
+
+use ScummVM\Objects\Company;
+
+/**
+ * The CompanysModel is used to cross reference companyies across the website
```
Thunderforge, line 7: Typo: `companyies` → `companies`

include/DataUtils.php
```diff
@@ -0,0 +1,48 @@
+<?php
+namespace ScummVM;
+
+require_once __DIR__ . '/../vendor/autoload.php';
+require_once __DIR__ . '/../include/Constants.php';
+
+use League\Csv\Reader;
+use League\Csv\Statement;
+
+class DataUtils
```
Thunderforge, line 10: I think a comment should be added explaining this class in detail, given that it generates the data that powers so much of this website.

data/versions.yaml
```diff
@@ -0,0 +1,129 @@
+# This is a generated file, please do not edit manually
```
Thunderforge, line 1: My preferred name for the file would; be `scummvm_versions.yaml` to clarify that it's for the app, not for any of the supported games.

data/versions.yaml
```diff
@@ -0,0 +1,129 @@
...
+---
+- id: 0.1.0
+  date: 1/13/2002
+  name: ""
+- id: 0.2.0
+  date: 5/14/2002
+  name: ""
+- id: 0.3.0
+  date: 12/1/2002
```
Thunderforge, line 10: I would prefer that dates be in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format. ScummVM has a global audience, and we shouldn't be storing in dates that could be confused by readers from different parts of the globe (is this December 1st or January 12th)?

data/versions.yaml
```diff
@@ -0,0 +1,129 @@
+# This is a generated file, please do not edit manually
+---
+- id: 0.1.0
+  date: 1/13/2002
+  name: ""
```
Thunderforge, line 5: It looks like all releases have this field blank. What's the purpose of this?

data/games.yaml
```diff
@@ -0,0 +1,3926 @@
+# This is a generated file, please do not edit manually
+---
+- id: martian
+  name: Martian Memorandum
+  engine_id: access
+  company_id: access
+  moby_id: "222"
```
Thunderforge, line 7: Why does the `moby_id` need quotes if other fields don't? It seems redundant and I would prefer to remove them if we can.

data/games.yaml
```diff
@@ -0,0 +1,3926 @@
...
+  datafiles: ""
+  series: ""
+- id: amazon
+  name: 'Amazon: Guardians of Eden'
```
Thunderforge, line 11: These single-quotes should be removed; Martian Memorandum above doesn't have them, so I assume they're optional.

data/games.yaml
```diff
@@ -0,0 +1,3926 @@
...
+  series: ""
+- id: dragons
+  name: Blazing Dragons
+  engine_id: dragons
+  company_id: crystaldynamics
+  moby_id: "8513"
+  datafiles: ""
+  series: ""
+- id: drascula
+  name: 'Dráscula: The Vampire Strikes Back'
```
Thunderforge, line 704: I assume it has been verified that this will display properly; we had an issue a while back where non-ASCII characters weren't.

Thunderforge (changes requested):
Despite my many comments (inevitable due to the number of files, I presume), I'm very impressed with this effort. Great work!

Most of my comments were related to documentation; I had a bit of confusion as a reviewer about the purpose of some things, and I assume that future maintainers who want to make changes will too. Fortunately, they should be easy to address. I didn't really have any technical concerns aside from handling of `DEV` versions.

Looking forward to seeing this merged!

data/platforms.yaml
```diff
@@ -0,0 +1,96 @@
+# This is a generated file, please do not edit manually
+---
+- id: 3do
+  name: 3DO
+  enum: kPlatform3DO
```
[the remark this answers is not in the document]
Mataniko: I used it to map to a data dump of stable/unstable entries - it should just be removed in the data

data/platforms.yaml
Mataniko: I don't feel strongly about it either way - but i think this matches better on what's on the actual game boxes that people look at, or how these games are historically referred to

data/companies.yaml
Mataniko: That's how PHP yaml emit works.

data/companies.yaml
Mataniko: this is outdated, data has been fixed upstream

data/compatibility.yaml
Mataniko: markdown bullet list

data/compatibility.yaml
Mataniko: That's impossible to represent in tabular data

data/compatibility.yaml
Mataniko: i had it originally as from_version - i agree

Mataniko: The only issue is that everything at 2.0.0 is not "since version" 2.0.0. This field is more "relevant for version XXX and up"

include/Pages/CompatibilityPage.php
Mataniko: Hasn't happened in 20 years, and if it ever does happen - we can fix this.

include/Models/LegacyCompatibilityModel.php
Mataniko: We need to keep DEV because otherwise we'll break URL handling for /DEV

data/games.yaml
Mataniko: my guess is pulling the data from sheets everything looks like a string so it gets parsed to a string - or maybe that was a side effect of having non-numeric values in the column

Mataniko:
Thanks for the feedback - there's no point in commenting on the YAML files since they are generated - we can't control the exact structure. If there are data errors they should be fixed in the spreadsheet for now and we can refresh the data. 

Can you please resolve the comments where appropriate?

data/versions.yaml
Mataniko: Will eventually do that - we don't use the dates at the moment.__

data/games.yaml
Mataniko: It's due to the colon which is a special character

data/games.yaml
Mataniko: Yes

include/Pages/CompatibilityPage.php
Mataniko: (Only reason we assign an actual number to DEV is version comparison)

data/compatibility.yaml
Thunderforge: Huh, I didn't realize we were putting markdown within a string. If it works, I'll accept it; I had thought it was a holdover from old data.

data/compatibility.yaml
Thunderforge: I'm a bit confused about what you're saying here. Can you rephrase what you mean?

My understanding is that we have an entry (`maniac`) and the `2.0.0` means that this is the version it was introduced in, so it's available in this revision and later. Is that not how this works?

data/compatibility.yaml
Thunderforge: Since you said earlier that the string is later being parsed as markdown, I'm less concerned about this. I had missed that bit of information.

include/Pages/CompatibilityPage.php
Thunderforge: Alright, well like I said, I won't hold up the PR for it if you don't think it's worth fixing.

include/Models/LegacyCompatibilityModel.php
Thunderforge: Per other comments: I won't hold up the PR over this, so marking as resolved.

data/versions.yaml
Thunderforge: If the dates aren't being used right now, then it's just a matter of changing them in the spreadsheet? Or do we need to make changes to this code so that if it is changed in the spreadsheet it won't break things?

data/platforms.yaml
Thunderforge: @sev- Maybe you can be a tiebreaker here? Should the platform be labeled with the historic name of `Macintosh` or the modern name (as used on Wikipedia) of `Classic Mac OS`?

include/DataUtils.php
Thunderforge: Thanks for adding the description.

Thunderforge (commented):
Had a few outstanding items, otherwise I'm glad to see that my concerns were addressed.

data/compatibility.yaml
Mataniko: I'm not planning to migrate the old compatibility lists to the new format due to the amount of work required with no real value. right now 99% of the list is "since version 2.0.0" which is not the version the game was introduced to scummvm, it's the version the compatibility entry is relevant since

data/versions.yaml
Mataniko: spreadsheet - i recently added support to automatically convert epoch dates to localized date strings, that's the format we should use

data/games.yaml
Mataniko: if it ain't broke... they're will be used in HTTP requests, and will have to be casted to a string at that point anyway.

data/platforms.yaml
Mataniko: We are referring to the computer platform here, not the operating system.

data/compatibility.yaml
Thunderforge: So if I understand this correctly then:

* All games have an entry corresponding with version 2.0.0
* Games with changes to their information since 2.0.0 will have another entry (say, 2.1.0) that will include any changes to the original 2.0.0 entry.

Is that correct?

data/versions.yaml
Thunderforge: Okay, glad to hear that it's something that can easily be changed in the spreadsheet without breaking code in this PR.

data/platforms.yaml
Thunderforge: I'm not sure I understand the distinction. We use operating systems for other entries like DOS, Windows, Linux, and AmigaOS.

Mataniko: The nomenclature changed over the years for platforms vs OS. When you bought a game it would say "DOS compatible" or "MS-DOS compatible" or "Macintosh compatible" 

If anything - AmigaOS is the wrong nomenclature.

data/compatibility.yaml
Mataniko: More accurately:
All games have an entry, which most of them correspond with 2.0.0 since that's the cutoff for the data conversion
Yes, changes will overwrite previous entries.

data/platforms.yaml
Thunderforge: I suppose it's something that can be change easily since it's just data and isn't directly related to the code you're merging. I'll resolve the conversation for now so it doesn't hold up merging this code.

Thunderforge (approved):

Mataniko:
The last few commits introduce caching to data models.

Today our website has to read from data from disk every time a page is served - this moves data to redis which should speed things up significantly and reduce IO

lotharsm (approved):

lotharsm:
Thanks for your work on this, @Mataniko! Should we merge this so further development can continue in-tree?

Mataniko:
I'm planning to merge this alongside 2.2.0 release. This is a pretty major change so i'm trying to find breakage now.

criezy:
I only skimmed the changes, but it looks like this is great work!
My only worry is that I suspect the change to the screenshot names and paths will break the screenshots on the game pages on the wiki (as they get their screenshots from the website as far as I am aware).

PR #132, opened 2020-09-12
WEB: Clarify descriptions in Contact page

SupSuper:
While translating the website I ran into some confusing sentences in Contact that I thought could be clarified, since it's an important part of the website:

> **Original**
Please do not contact the team for questions about using ScummVM. Instead use the [forums](http://forums.scummvm.org/) or the [bug reporting system](http://bugs.scummvm.org/)!
**Replacement**
Please check the [documentation](/documentation/) and [wiki](http://wiki.scummvm.org/) before asking questions about using ScummVM!

I get the intent of the original sentence "don't bother the team with trivial matters". But then you just link to contacts already described below, which often contradict this line! The bug tracker is not for support issues, the forum is used both by the community and team, etc. So it feels redundant and confusing.
Instead I replaced it with a call to RTFM. :)

>  **Original**
There are three ScummVM related [mailing lists](http://lists.scummvm.org/). Two of them are for automated content only. The one where you can send emails to yourself is scummvm-devel.
**Replacement**
There are three ScummVM related [mailing lists](http://lists.scummvm.org/). You can only send emails to scummvm-devel, the rest are for automated notifications only.

The original reads like the mailing list is for talking to yourself (which might be these days since everyone moved to other platforms :)) so I clarified that.


criezy:
I support both changes.

data/localization/en_US/strings.json
```diff
@@ -26,7 +26,7 @@
   "compatibilitySectionTitle": "{company} Game Compatibility Chart",
   "contactTitle": "Contact",
   "contactContentTitle": "Contact",
-  "contactIntro": "**Please do not contact the team for questions about using ScummVM. Instead use the [forums](http://forums.scummvm.org/) or the [bug reporting system](http://bugs.scummvm.org/)!**",
+  "contactIntro": "**Please check the [documentation](/documentation/) and [wiki](http://wiki.scummvm.org/) before asking questions about using ScummVM!**",
```
Mataniko, line 29: Might as well just change it to reflect whats' on the page: "You can reach us via the following methods:" or something in that manner

Thunderforge: I'm fine with either the current version or further changes as suggested by Mataniko.

SupSuper:
Updated

Thunderforge (approved):

Thunderforge:
@SupSuper Do you have the ability to squash and merge commits to this repo? If not, I can do that.

PR #134, opened 2020-09-29
Drop lang query param and cookie and move to canonical URLs for translations.

Mataniko:
This PR reverts back some of the previous changes and migrates away from using the lang query param and cookie to set the site language.

Instead, we now use a canonical URL path, along side alt links in the document head. 

This has a few benefits:

- Compliant with Google's indexer: https://support.google.com/webmasters/answer/189077?hl=en
- Canonical URLs to any language
- Users who check back and their cookie expired do not get surprised with a different language.
- Easier localization of other data files in the future

Cons: 
- Slightly more complex markup in templates (which don't change often)

I tried to preserve some backwards compatibility such as redirecting users coming in with a query param or cookie.

Demo available at https://staging.scummvm.org/


sev-:
looks nice

PR #135, opened 2020-09-30
WEB: Update URLs in Links page

SupSuper:
This updates all the URLs in the Links page to the latest versions / HTTPS.

Some other things I noticed:
- The links in "Technical information about SCUMM" are gone, should we remove this section?
- Should we remove the Ultima engine links now that they've merged with ScummVM?

### DIFF
--- a/data/links.yaml
+++ b/data/links.yaml
@@ -7,12 +7,12 @@
       description: >-
         SDL (Simple DirectMedia Layer) is a cross-platform multimedia library,
         and used by the primary backend of ScummVM.
-      url: 'http://www.libsdl.org/'
+      url: 'https://www.libsdl.org/'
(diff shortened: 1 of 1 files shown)


Thunderforge (approved):

Thunderforge:
> The links in "Technical information about SCUMM" are gone, should we remove this section?

You mean they are dead links? Yeah, we should remove it if that's the case. It's too big of a reference to link to the Wayback machine, I think, and is becoming more niche as we expand to more engines. It would be better suited for the wikis. I'm in favor of small PRs, so feel free to create a new PR for that.

> Should we remove the Ultima engine links now that they've merged with ScummVM?

I'm in favor of this in a new PR as well.

PR #138, opened 2020-10-05
WEB: Putting all compatibility data into one sortable table (#110)

### LINKED ISSUE #110: Combine game compatibility list into one table with sortable columns

Thunderforge:
Closes #110. This PR does the following:

- Merges all game compatibility info into one table
- Adds the ability to sort table data alphabetically by column (via [tablesort.js](https://github.com/tristen/tablesort)). Default sort is by game full name.
- Removes the company name from the caption
- Adds the "Company" column

There are still a couple of tasks that need to be completed before this PR can be merged:

- [ ] Modify the data structure of compatibility data to be more efficient
- [ ] Localize all modified strings

Full Name Sort

Company Sort Ascending

Company Sort Descending


data/localization/en/strings.json
```diff
@@ -17,6 +17,7 @@
   "compatibilityDetailsChartCol2": "Game Short Name",
   "compatibilityDetailsChartCol4": "Support Level",
   "compatibilityDetailsChartCol3": "Datafiles",
+  "compatibilityDetailsChartColCompany": "Company",
```
Thunderforge, line 20: I think that it would be better to give these strings keys based on their function instead of by their index, as we may decide to rearrange, add, or remove columns in the future.

templates/pages/compatibility.tpl
```diff
@@ -54,37 +55,43 @@
...
-            <td class="gameDatafiles">
-                {if $game->getDatafiles()}
-                <a href="{$game->getDatafiles()}">{#compatabilityDetailsDetails#}</a></td>
-            {else}
-            ---
-            {/if}
-            <td class="gameSupportLevel {($game->getVersion() == $version) ? ' updated' : $pct_class}">{$support_level}</td>
-        </tr>
+        {foreach from=$compat_data key=company item=games name=compat_loop}
+            {foreach from=$games item=game}
```
Thunderforge, line 71: This was just a quick hack to get it working. As shown in the checklist on the initial PR comment, we will want to simplify the data structure for compatibility data.

Mataniko:
I took a look at the current state and there's definitely work to do here.

1. Having everything in one table makes it difficult to navigate the data, even with sorting
2. The current tables are already sorted by name and grouped by company - there's not much value of sorting by game id (and none by data files) - leaving only support level to have some merit in sorting

A few things that can help:
1. Grouping
2. Filtering
3. Moving away from a table to responsive markup

Thunderforge:
> Having everything in one table makes it difficult to navigate the data, even with sorting

@Mataniko Is it more difficult than the current method? Right now if I know the name of the game, but not the company (perhaps for a lesser known game like "Once Upon A Time: Little Red Riding Hood"), I have to look through each company table (and the "Other" table) to find what I'm looking for.

My goal is to do no harm and not make it harder than the current method. If you think that this is more difficult than what we currently have, I can reconsider.

> The current tables are already sorted by name and grouped by company - there's not much value of sorting by game id (and none by data files) - leaving only support level to have some merit in sorting

I agree. I would love to add columns for engine and series, both of which would benefit from sorting, and move the data files and maybe the short name to the drilldown page. My thought was to have this PR be smaller and those changes added later, but I can do them if desired.

As for grouping, filtering, and responsive markup, I agree they should be added, but do you think they should be part of this PR? In other words, do we want smaller PRs that incrementally make improvements or larger PRs that completely reach our desired goal?

Mataniko:
The "know the name of the game but not the company" is easily solvable by a filter or ctrl+f. We don't have any metrics on how users use the table so i can't tell you with any degree of confidence what's easier - but we haven't had complaints about the current structure.

Adding columns would also require re-thinking the whole components, since the table stops working in smaller resolutions. 

I don't think it's worth regressing the current state in order to introduce incremental changes at an unknown time in the future.

Thunderforge:
Closing due to age. May revisit in the future.

bluegr:
The tablesort part of this PR is interesting and could be merged separately

PR #156, opened 2021-08-25
Changing table hyperlinks from black to default color

Thunderforge:
This makes hyperlinks consistent with other pages, including ones with lots of links such as the Downloads page

## Before

Clickable Compatibility Table
Clickable Game
Clickable Demo

## After

image
image
image

The concern was raised in Discord that there was too much red now on the tables. For comparison, we already have a similar amount of red on the downloads page, so this change is consistent with the rest of the site.

image

Since the amount of red on that page is considered acceptable, I do not think that this is too much here either, and it helps with website consistency. Plus, red hyperlinks means the clickable words can now be easily distinguished from the unclickable words.

### DIFF
--- a/scss/theme/_themes.scss
+++ b/scss/theme/_themes.scss
@@ -38,7 +38,6 @@ $themes: (
         chart-row-even: $white,
         chart-caption-background: $totem-pole,
         chart-caption-header: $white,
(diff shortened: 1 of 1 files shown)


sev-:
Thanks!

PR #161, opened 2021-09-01
WEB: Add a Wikipedia link for games

Thunderforge:
Add a new field to the game table called `wikipedia_page`, which is the URL page title for the game's corresponding page on English Wikipedia.

The field has been filed out for a few games (such as *Martian Memorandum* and *Manhunter: New York*), but more will need to be filled out.

image

### DIFF
--- a/include/OrmObjects/Compatibility.php
+++ b/include/OrmObjects/Compatibility.php
@@ -54,6 +54,9 @@ class Compatibility extends BaseCompatibility
         if ($this->getGame()->getDataFiles()) {
             $links[] = "- [ScummVM Wiki]({$this->getGame()->getDataFiles()})";
         }
+        if ($this->getGame()->getWikipediaPage()) {
+            $links[] = "- [Wikipedia](https://en.wikipedia.org/wiki/{$this->getGame()->getWikipediaPage()})";
+        }
         if ($links) {
             $notes .= "\n\n**Links:**\n";
             $notes .= join("\n", $links);
--- a/include/OrmObjects/Compatibility.php
+++ b/include/OrmObjects/Compatibility.php
@@ -54,8 +54,15 @@ class Compatibility extends BaseCompatibility
         if ($this->getGame()->getDataFiles()) {
             $links[] = "- [ScummVM Wiki]({$this->getGame()->getDataFiles()})";
         }
-        if ($this->getGame()->getWikipediaPage()) {
+        $wikipediaPage = $this->getGame()->getWikipediaPage();
+        if ($wikipediaPage) {
+          // If we have a full URL, such as for a non-English Wikipedia page, use that
+          // Otherwise, assume it's a page for English Wikipedia
+          if (str_starts_with($wikipediaPage, "https://")) {
+            $links[] = "- [Wikipedia]({$this->getGame()->getWikipediaPage()})";
+          } else {
             $links[] = "- [Wikipedia](https://en.wikipedia.org/wiki/{$this->getGame()->getWikipediaPage()})";
+          }
         }
         if ($links) {
             $notes .= "\n\n**Links:**\n";
--- a/schema.xml
+++ b/schema.xml
@@ -10,6 +10,7 @@
     <column name="engine_id" type="varchar" size="24" required="true"/>
     <column name="company_id" type="varchar" size="24" required="true"/>
     <column name="moby_id" type="integer"/>
+    <column name="wikipedia_page" type="varchar"/>
     <column name="series_id" type="varchar"/>
     <foreign-key foreignTable="engine" onDelete="CASCADE">
       <reference local="engine_id" foreign="id"/>


data/en/games.yaml
```diff
@@ -4662,6 +5244,7 @@
     company_id: adeline
     moby_id: '748'
     datafiles: 'https://wiki.scummvm.org/index.php?title=Little_Big_Adventure'
+    wikipedia_page: ''
```
mgerhardy, line 5247: This would be `Little_Big_Adventure` ;)

https://en.wikipedia.org/wiki/Little_Big_Adventure

data/en/games.yaml
```diff
@@ -4670,6 +5253,7 @@
     company_id: adeline
     moby_id: '1765'
     datafiles: ''
+    wikipedia_page: ''
```
mgerhardy, line 5256: and here `Little_Big_Adventure_2` - https://en.wikipedia.org/wiki/Little_Big_Adventure_2

sev-:
Nice idea and addition, however, for easier development, I would recommend adding a possibility to add full URL to Wikipedia, as then there is much less copy/pasting. That will also give a possibility to specify non-English main Wikipedia page for the cases when some game doesn't have an English page.

E.g. the logic is something like:

```
  if (!address.hasPrefix("https://"))
     address = "https://en.wikipedia.org/wiki/" + address;
```

Thunderforge:
You raise a good point about non-English Wikipedia pages. Since our ScummVM website is multilingual, what if instead we had fields for multiple Wikipedia language pages?

For instance, say we had the following fields in the spreadsheet:
* `wikipedia_en`: https://en.wikipedia.org/wiki/The_Secret_of_Monkey_Island
* `wikipedia_de`: https://de.wikipedia.org/wiki/The_Secret_of_Monkey_Island

Then on our Compatibility pages
* [English page for The Secret of Monkey Island](https://www.scummvm.org/en/compatibility/DEV/monkey/) shows a link to "[Wikipedia (English)](https://en.wikipedia.org/wiki/The_Secret_of_Monkey_Island)"
* [German page for The Secret of Monkey Island](https://www.scummvm.org/de/compatibility/DEV/monkey/) shows a link to "[Wikipedia (Deutsche)](https://de.wikipedia.org/wiki/The_Secret_of_Monkey_Island)", or instead  "[Wikipedia (Englische)](https://en.wikipedia.org/wiki/The_Secret_of_Monkey_Island)" if `wikipedia_de` is blank, but `wikipedia_en` exists.

We could create some order of precedence for which ones display if there isn't a Wikipedia page in the current language.

data/en/games.yaml
Thunderforge: There are hundreds of games that do have a Wikipedia page that we could add via [this spreadsheet](https://docs.google.com/spreadsheets/d/1QzwFleEKXOsE59cYMOcQB7C2f0Np48uAQOCG8kicX_s/edit?usp=sharing) (which gets automatically turned into the YAML files that are checked in). The goal of this PR is to prove that a Wikipedia link can be added, and to that end I filled out the field for only a couple of games. 

Once we decide on the correct format for adding links (it might change; see other comments in this PR), we can add all the Wikipedia pages. We'll probably have a script go and find the hundreds of Wikipedia pages.

data/en/games.yaml
Thunderforge: See [this](https://github.com/scummvm/scummvm-web/pull/161#discussion_r700384384).

sev-:
IIRC, there was some Wikipedia API that redirects to a given page from an English URL. I can check further, but I remember it being used by OpenStreetMap. Adding the language columns to the spreadsheet seems to me as an overkill and too much of data to handle (as we require filling out the spreadsheet when a game is announced supported).

Thunderforge:
An API that redirects to the browser's language would definitely be an ideal solution. Less data for us while allowing multilingual support.

I did some digging and found https://qrpedia.org/. It's designed for QR codes, but fortunately it does URL redirection too. From some documentation

> For example: https://en.wikipedia.org/wiki/Cat, if changed to https://en.qrwp.org/Cat, will be redirected in a french browser to https://fr.m.wikipedia.org/wiki/Chat.

Unfortunately, it only redirects to mobile pages, even if you're on a desktop. I couldn't find any way to overcome that.

The only other alternative I found was [this page on Wikidata](https://www.wikidata.org/wiki/Special:GoToLinkedPage?site=&itemid=Q145568). If you type in `frwiki` to the site field, or add it to the URL, then it will go to the appropriate site. But there isn't any way to automatically detect the browser's language, and if you try to reference a site that doesn't have an article (like trying to [find a ScummVM page on the Esperanto Wikipedia](https://www.wikidata.org/wiki/Special:GoToLinkedPage?site=eowiki&itemid=Q145568)), then you get thrown back to the Wikidata page, which is not the user experience we want.

@sev- Were you able to determine if there was another API that OpenStreetMap used?

criezy:
Since Wikipedia is using the mediawiki software, the mediawiki API is available.
https://mediawiki.org/wiki/API:Main_page

For example
```
https://en.wikipedia.org/w/api.php?action=query&format=json&prop=langlinks&titles=Little%20Big%20Adventure&llprop=url&lllang=fr
```
returns
```json
{
    "batchcomplete": "",
    "query": {
        "pages": {
            "1017834": {
                "pageid": 1017834,
                "ns": 0,
                "title": "Little Big Adventure",
                "langlinks": [
                    {
                        "lang": "fr",
                        "url": "https://fr.wikipedia.org/wiki/Little_Big_Adventure",
                        "*": "Little Big Adventure"
                    }
                ]
            }
        }
    }
}
```

Thunderforge:
It appears to me that the Media Wiki software still requires inputting the language, meaning the burden is on us to figure out the user's language. Also, if the page doesn't exist, we don't get something that the user can be directed to ([example](https://en.wikipedia.org/w/api.php?action=query&format=json&prop=langlinks&titles=ScummVM&llprop=url&lllang=eo)).

Thunderforge:
@sev- I implemented your suggestion to allow for a full Wikipedia link if provided, or else use a page stub and assume it's for the English Wikipedia.

sev-:
Excellent, let's merge this! How now to fill out the missing pages? Shall it be done manually or there is a way to automate the initial seed somehow?

sev-:
And btw, the code bitrot, please, rebase it.

Thunderforge:
> Excellent, let's merge this! How now to fill out the missing pages? Shall it be done manually or there is a way to automate the initial seed somehow?

@sev- Most of the `wikipedia_page` entries have already been added on the spreadsheet, and it got merged into this repo  with c54e0030d0b38698205a0fd7f9d081b6b96a71cb. Not sure who added them, but I'm glad they did!

> And btw, the code bitrot, please, rebase it.

Done!

Thunderforge:
[The new Codacy issue](https://app.codacy.com/gh/scummvm/scummvm-web/file/60782227504/issues/source?bid=21552110&fileBranchId=25225428#l63) seems to want me to remove the else in if/else. But the fix instructions say to do an early return, which I can't do because the if/else is in the middle of the method. How does it want me to fix it?

criezy:
> [The new Codacy issue](https://app.codacy.com/gh/scummvm/scummvm-web/file/60782227504/issues/source?bid=21552110&fileBranchId=25225428#l63) seems to want me to remove the else in if/else. But the fix instructions say to do an early return, which I can't do because the if/else is in the middle of the method. How does it want me to fix it?

That codacy issue seems a bit stupid to me. I suspect it did not realise there was more code after the if/else block. I suggest you just ignore it.

sev-:
Weird, I see this:

Screenshot 2021-09-08 at 15 03 14

Thunderforge:
For any onlookers: the issue was apparently that I had merge commits, whereas Sev told me that the repo requires linear commits. Not sure why that resulted in us seeing different things, but at least once I properly rebased, we both saw there were no conflicts.

PR #162, opened 2021-09-07
WEB: Fixing issue redirecting games with hyphens

Thunderforge:
Fixing an issue that prevented loading the compatibility page for games with hyphens in their short name, such as `private-eye` and `agi-fanmade`. Previously, attempting to go to the link for such a page would result in redirecting to the index page. Now the page is properly loaded.


sev-:
Thanks!

PR #163, opened 2021-09-20
DATA: Removing redundant notes on supported platforms

Thunderforge:
For instance, `lzone` has `stable_platforms: 'win,mac'`, so it is redundant to have notes of

> Only Windows and Mac versions are supported

## Before
Redundant Notes

## After
No Notes

### DIFF
--- a/data/en/compatibility.yaml
+++ b/data/en/compatibility.yaml
@@ -3754,7 +3754,7 @@
 -
     id: chest
     support: good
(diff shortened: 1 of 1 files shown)


lotharsm:
Thank you! I confirmed that our datasheet contains the correct entries. Since the description on the websites is dynamically generated, I re-triggered the website generation, so your changes are live.

Closing this due to the auto generation.

PR #164, opened 2021-09-21
WEB: Changing "Daily Snapshots" link to "Daily Builds"

Thunderforge:
The URL goes to [a section](https://www.scummvm.org/downloads/#daily) that has "Daily Builds" as its header, so the link name should be consistent.

## Before

image

## After

image


lotharsm:
Thanks, merging!

PR #165, opened 2021-09-29
DATA: Removing other game engines from links

Thunderforge:
All the Ultima engines have been merged with ScummVM. That leaves DOSBox, which has a description that says "DOSBox does not really quite fit in here", which means it's not really worth keeping.

### DIFF
--- a/data/en/links.yaml
+++ b/data/en/links.yaml
@@ -80,39 +80,6 @@
         The personal blog of Ron Gilbert, the man behind many great LucasArts
         adventures, like Maniac Mansion and Monkey Island.
       url: 'https://grumpygamer.com/'
-- name: Other classic game engine open source projects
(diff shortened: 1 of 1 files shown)


criezy:
I don't think Exult (for Ultima VII) has been merged with ScummVM.

Thunderforge:
You're right, looking more closely, I don't see any indication on the web page has been merged. 

xu4 (Ultima IV) has a news item saying it's now in ScummVM while Nuvie (Ultima VI et al) and Pentagram (Ultima VII et al) have ScummVM in their logos.

So that means we either keep just Exult (Ultima VII) and/or DOSBox, or delete them all. Or add some more, I suppose, but where do we stop?

Thunderforge:
Discord consensus:

image

PR #166, opened 2021-09-29
WEB: Updating composer.lock and package-lock.json

Thunderforge:
Looks like these haven't been updated in a long time. I updated them this morning and this is what I got.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "altorouter/altorouter",
-            "version": "2.0.1",
+            "version": "2.0.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/dannyvankooten/AltoRouter.git",
-                "reference": "127f6e96998708a31ef32252985bea82e3b03888"
+                "reference": "f6fede4f94ced7c22ba63a9b8af0bf2dc38e3cb2"
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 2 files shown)


Thunderforge:
@lotharsm You are correct that we should be modifying `package.json` files. When you run `composer develop` (or one of the related tasks), that will also update the `package-lock.json` files. [These are intended to be checked in](https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5).

[Here is the documentation on how to deal with lockfile conflicts](https://docs.npmjs.com/cli/v6/configuring-npm/package-locks#resolving-lockfile-conflicts). I've not actually tried any of these methods, but that's what they recommend.

> since the deployment scripts re-writes the package-lock.json file on it's own instead of fetching it from git.

I'm wondering if that's the issue. My understanding is that the intended use case is that `package-lock.json` is pulled down and won't be modified so long as all the versions are still compatible. 

For instance, if `package.json` is set to `~1.2.3` (i.e. `>= 1.2.3 and < 1.3.0`) and `package-lock.json` file says that `1.2.4` was pulled in (which matches the ``~1.2.3` criteria), then the CI server should automatically be suing `1.2.4` even if `1.2.5` is available. Thus, no changes are needed to `package-lock.json`

If the build scripts are doing something to update the dependencies or something, then we will have a conflict. But the CI shouldn't be updating dependencies anyway; that should be a dev process.

(Caveat emptor: I am not an expert at web programming, so I may have a misunderstanding of how this all works).

PR #168, opened 2021-10-18
WEB: Changing headers to use Markdown headers instead of bold

Thunderforge:
## Before

image

## After

image


lotharsm:
Thank you - this is a nice way to improve consistency. Merging.

PR #172, opened 2021-10-30
WEB: Removing file.getType() and related dead code

Thunderforge:
It appears that at one point the data spreadsheet included a `type` field. This is no longer the case and so all of this is not used, providing unnecessary confusion and maintenance.

### DIFF
--- a/templates/components/list_items.tpl
+++ b/templates/components/list_items.tpl
@@ -10,26 +10,12 @@
                         <span class="download-extras">
                             {if is_array($data)}
                                 (
-                                {if $item->getType() == 'daily'}{#listItemsBuildFromRepo#} {/if}
                                 {$data.size} {if $data.ext == '.exe'}Win32 {/if}{$data.ext}{if $data.date != ""}{#listItemsDate#} {$data.date} {/if}
                                 &nbsp;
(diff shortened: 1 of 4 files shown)


templates/components/list_items.tpl
```diff
@@ -10,26 +10,12 @@
                         <span class="download-extras">
                             {if is_array($data)}
                                 (
-                                {if $item->getType() == 'daily'}{#listItemsBuildFromRepo#} {/if}
```
Thunderforge, line 13: Will`listItemsBuildFromRepo` be automatically removed from the localization files? Or do I need to go through `en/strings.json`, `de/strings.json`, etc and remove them all?

lotharsm: You need to modify `en/strings.json` as well. The other language files should remain untouched since the translation system will pick up the necessary changes.

lotharsm: Following the discussion on Discord: I recommend you modify `en/strings.json` first, then I'll merge this PR - we'll see if the translation system is as smart as I remember ;)

Thunderforge: Done, thanks!

lotharsm:
Thank you, merging!

PR #174, opened 2021-11-08
DATA: Changing downloads table to use version column

Thunderforge:
In order to simplify the process of updating version downloads, [downloads spreadsheet table](https://docs.google.com/spreadsheets/d/1QzwFleEKXOsE59cYMOcQB7C2f0Np48uAQOCG8kicX_s/edit#gid=1057392663) now includes a `version` column and the `url` field is much simpler. In order to update a new version, simply change the value in the `version` field.

Further improvements can be made, but for now this is an improvement that keeps parity with the previous code.

----

When this code gets merged, the following steps should be taken on the spreadsheet:

1. Delete the old `scummvm_downloads` table
2. Rename `Thunderforge_wip_scummvm_downloads` to `scummvm_downloads`

Because the code references sheet ids and not tab names, no further action needs to be taken.

### DIFF
--- a/include/Constants.php
+++ b/include/Constants.php
@@ -47,7 +47,7 @@ class Constants
 
         /* Downloads */
         define('DOWNLOADS_BASE', 'https://downloads.scummvm.org');
-        define('DOWNLOADS_URL', '/frs/scummvm/{$release}/');
+        define('DOWNLOADS_URL', '/frs/scummvm/{$version}/');
         define('DOWNLOADS_DAILY_URL', '/frs/daily/');
         define('DOWNLOADS_TOOLS_URL', '/frs/scummvm-tools/{$release_tools}/');
         define('DOWNLOADS_EXTRAS_URL', '/frs/extras/');
--- a/include/DataUtils.php
+++ b/include/DataUtils.php
@@ -31,7 +31,7 @@ class DataUtils
         'game_demos' => '713475305',
         'series' => '1095671818',
         'screenshots' => '1985243204',
-        'scummvm_downloads' => '373699606',
+        'scummvm_downloads' => '1057392663',
         'game_downloads' => '1287892109',
         'director_demos' => '1256563740',
     ];
--- a/include/Models/DownloadsModel.php
+++ b/include/Models/DownloadsModel.php
@@ -85,11 +85,15 @@ class DownloadsModel extends BasicModel
             ->setIgnoreCase(true)
             ->findByUserAgent($os['name']);
 
-        foreach ($downloads as $download) {
-            $url = str_replace('{$release}', RELEASE, $download->getURL());
-            sscanf($url, "/frs/scummvm/%s", $versionStr);
-            $version = substr($versionStr, 0, strpos($versionStr, "/"));
+        if (!empty($downloads)) {
+            $download = $downloads[0];
+
             $name = strip_tags($download->getName());
+            // Construct the URL and fill in the version
+            $url = DOWNLOADS_BASE . DOWNLOADS_URL . $download->getURL();
+            $version = $download->getVersion();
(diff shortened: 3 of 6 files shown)


sev-:
Nice thing! Two questions, though:
* Could the category column be dropped and the version field used for it
* Could the current/legacy files be mixed with keeping their relative order?

Thunderforge:
> Could the category column be dropped and the version field used for it

@sev- Right now, no. I would like to make that change, but right now two columns are needed to support our nested hierarchy.
image
I think it's worth exploring if we want to change the structure, but it's probably best at another PR, since this one is focused on under the hood changes.

> Could the current/legacy files be mixed with keeping their relative order?

Yes, legacy/current ordering doesn't matter. However, the first listed in a category will be displayed first. This is why Windows is at the top of the list instead of near the bottom.

sev-:
> > Could the category column be dropped and the version field used for it
> 
> @sev- Right now, no. I would like to make that change, but right now two columns are needed to support our nested hierarchy. 

Could you please explain a bit? I do not understand. I am referring to the distinction between current/legacy.

So, potentially, you merge those two together into a new category 'scummvm', but then is it current or legacy will be determined by the version number.

Also, regarding the order, I meant something like this:

```
  TRUE | scummvm | 2.5.0 | windows build
  TRUE | scummvm | 2.3.0 | PS2 build
  TRUE | scummvm | 2.5.0 | linux build
```

which then will be split into two lists with keeping their order, e.g.

```
  current:  windows, then linux
  legacy: ps2
```

This way, I will be flipping only the version numbers and the relative positions of every platform will stay the same.

Thunderforge:
You are correct that we could find a way to do what you are describing there: one column with a category (with a value called `scummvm` for example) and then the version determines which sub-category it goes into.

The problem is with the other stuff on the downloads page. We have

* Extras
   * Games
   * Subprojects
* Libs
   * Required
   * Optional

I don't see how we can keep that category-subcategory relationship with one column.

I am open to redesigning the page so that we no longer have subcategories in those, but I think that's best for a second PR. Smaller changes per PR and all.

sev-:
Looks like there is a misunderstanding. I am not asking to _drop_ the column anymore, but change its values.

So, for ScummVM binaries, you have these two pairs of values: "current/release" and "legacy/old". I propose to merge those two and use "version" column for determining is it current or legacy. Also, I would like the values to be in a fixed order, merged together, so I do not need to move old releases down when certain builds are not yet submitted.

Thunderforge:
> I propose to merge those two and use the "version" column for determining if it is current or legacy.

I understand this request. You want

```
category    | version
----------- | --------
scummvm     |  2.5.0    <-- appears under current section
scummvm     |  2.4.0    <-- appears under outdated section
```

I think that this is a good idea, but I don't know how to do this and *also* have the single column express the hierarchy of 

* Extras
    * Games
    * Subprojects
* Libs
    * Required
    * Optional

These downloads are currently on lines 175 and later.

> Also, I would like the rows to be in a fixed order, merged together, so I do not need to move old releases down when certain builds are not yet submitted.

That is possible in its current state. It should be possible to have something like

* Windows - 2.5.0
* Dreamcast - 2.4.0
* macOS - 2.5.0
* Wii - 2.4.0

And then Windows & macOS would show up in that order under Current and Dreamcast & Wii would show up in that order under the outdated.

If you want me to change the order on the spreadsheet, I would just want to know the order you want them to be in. Alphabetical? Most popular (do we have stats for that)? Something else?

Also, I would really prefer to merge this PR as-is and then I can submit future PRs if we want to make further changes. Smaller PRs mean less risk of unintended side effects. It would also mean that we no longer have to maintain two versions of the spreadsheet tab.

Thunderforge:
@sev- looked over this and told me over Discord that it was ready to merge.

PR #180, opened 2021-11-19
I10N: Removing Polish translation

Thunderforge:
This translation has 0% of its strings translated, meaning it's pointless to include. Removing until such a time as someone actually wants to translate it.

The [corresponding Weblate translation page](https://translations.scummvm.org/projects/scummvm-website/strings/pl/) should be removed too so that the files don't get regenerated.

## Before
Polish Before

## After
Polish After


lotharsm:
Thank you! Merging.

PR #183, opened 2021-11-25
WEB: Sorting downloads by version after a new one is inserted

Thunderforge:
Previously, downloads always got added to the end of the list. Now, the download is added to the end and then the list is sorted by version, meaning that the download will be moved to be the last one of all the downloads with that version.

## Before

Version 0.11.1 was at the bottom of the list.

image

## After

Version 0.11.1 is now properly after 1.0.0, but before 0.9.1.

image


sev-:
cool!

PR #184, opened 2021-11-28
I10N: Changing translations link to general project page

Thunderforge:
There are several translation projects aside from the ScummVM, such as the website and several game translations. Instead of linking to only the ScummVM project, now I link to the overall project that shows them all.

## Before

A two-line hyperlink labeled "Translate ScummVM" goes to https://translations.scummvm.org/projects/scummvm/scummvm/

image

## After

A one-line hyperlink labeled "Translations" goes to https://translations.scummvm.org/projects/

image

### DIFF
--- a/data/en/menus.yaml
+++ b/data/en/menus.yaml
@@ -34,7 +34,7 @@
     - name: '{#menuDevelopmentPullRequests#}'
       href: 'https://github.com/scummvm/scummvm/pulls'
     - name: '{#menuDevelopmentTranslations#}'
-      href: 'https://translations.scummvm.org/projects/scummvm/scummvm/'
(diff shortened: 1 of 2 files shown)


lotharsm:
Looks good to me. As you said, the translation system is not restricted to ScummVM itself anymore, so we should "expose" the game translations to the public as well to raise awareness.

However, modifying the translated files in `data` is not correct. Please leave them untouched, Weblate will mark them as "fuzzy" and ask the translators to update those strings.

Thunderforge:
@lotharsm Fixed. Now only the `data/en/strings.json` file is modified. We will expect Weblate to take care fo the rest.

PR #185, opened 2021-12-28
WEB: Removing collapsible demo lists

Thunderforge:
The current system of collapsible demo lists has a number of problems:

* It is inconvenient to use on mobile
* It hinders searching for text on the page (e.g. Ctrl+F and searching for "French" will not find all the French demos)
* The first demo (e.g. English Amiga) may have nothing to do with the other demos (e.g. French DOS), yet it is the only one visible when the group is collapsed

Therefore, I am removing the entirety of the collapsible lists to avoid these issues. The main downside is that the demo pages are now about double the length.

Perhaps we could find a new alternative method of organizing that does not suffer from these problems. Until then, this will allow everyone on mobile and desktop to easily find all the demos they might be looking for.

## Before

image

## After

image


bluegr:
The collapsible demo feature came up as a way to reduce the length of the demo page and make it easier to navigate. We haven't come up with another way to organize this page.

Perhaps the collapsible list feature can be toggled instead of removing it altogether?

lotharsm:
I think removing the "collapsible" feature even _improves_ visibility and how we can navigate the page. Due to the lack of a search function on the website, with the "always expanded" list, we can simply use the browser's search capability.

So _if_ we keep it, I think we should show the "full" page by default.

Thunderforge:
If we had the collapsed menus expanded by default, then I don't think anybody would ever collapse them, even if we had a "collapse all" button.

The reason that someone comes to the page is to find demos, especially ones in a language they understand. If a collapsible list is collapsed by default, that hinders them finding those demos. If it is expanded by default, then collapsing it does not help their objective.

I do think that there are better ways we can organize this page (maybe dropdown menus for language and platform?) but those seem like future PRs to be done after removing the existing functionality.

bluegr:
> I do think that there are better ways we can organize this page (maybe dropdown menus for language and platform?) but those seem like future PRs to be done after removing the existing functionality.

This is too vague. We should first have a suggestion before dropping the existing functionality.

Thunderforge:
Closing this for now because it sounds like there's disagreement on what the next steps are.

PR #186, opened 2022-01-09
WEB: Fixing extra info not appearing for ScummVM download links

Thunderforge:
## Before

No extra file information appears after hyperlinks.

Screen Shot 2022-01-08 at 6 50 54 PM

## After

Several pieces of file information appear after hyperlinks. Clicking on `sha256` results in the file hash appearing.

Screen Shot 2022-01-08 at 6 50 45 PM

### DIFF
--- a/include/Objects/File.php
+++ b/include/Objects/File.php
@@ -15,7 +15,6 @@ class File extends BasicObject
     {
         parent::__construct($data);
         $this->category_icon = $data['category_icon'];
-        $this->extra_info = $data['extra_info'] ?? null;
         $this->subcategory = $data['subcategory'] ?? null;
         $this->user_agent = isset($data["user_agent"]) ? $data["user_agent"] : "";
         $this->version = strtolower($data['version'] ?? null);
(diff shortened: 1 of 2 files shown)


include/Objects/File.php
```diff
@@ -15,7 +15,6 @@ public function __construct($data, $baseUrl = null)
     {
         parent::__construct($data);
         $this->category_icon = $data['category_icon'];
-        $this->extra_info = $data['extra_info'] ?? null;
```
Thunderforge, line 18: `extra_info` does not appear anywhere on the data sheet, so I've removed it (and the associated code checking for it below).

include/Objects/File.php
```diff
@@ -80,22 +79,10 @@ public function __construct($data, $baseUrl = null)
...
-            $this->url = $fname;
-        }
-        /**
-         * Get the filesize/last modified information and put it in
-         * $this->extra_info.
-         */
-        if (isset($attributes['extra_info'])
-            && $attributes['extra_info'] == 'true') {
-            if (is_file($fname) && is_readable($fname)) {
                 $this->extra_info['date'] = date('F j, Y, g:i a', @filemtime($fname));
```
Thunderforge, line 82: It's a bit messy in the code diff, but I moved the code setting the date information outside of an if-statement that always evaluated false into the same if-statement that resulted in checking for file size, etc. This means that the date field never used to show up, but now it does.

## Before
image

([From wayback machine](https://web.archive.org/web/20210624023748/https://www.scummvm.org/downloads/?p=downloads))

## After
Screen Shot 2022-01-08 at 6 50 45 PM

Thunderforge:
Right now, this only affects the [Downloads page](https://www.scummvm.org/downloads/). Additional refactoring on the data spreadsheet similar to what was done in #174 needs to be done before it will show up on the demos and games pages.

include/Objects/File.php
lotharsm: IIRC, the `extra_info` field was only used for the daily builds for Win32 (previously provided by Kirben and now by myself), and it was basically just telling who provided those builds.

Since they are now "official", I think removing the now unused field is a good idea since we are able to handle everything else in the datasheet.

include/Objects/File.php
lotharsm: Thank you for fixing this as well - I don't remember when the date field even worked for the last time besides of the daily builds. I think this has been broken for _years_ now.

lotharsm:
All in all, very nice work - thank you! Merging.

PR #187, opened 2022-01-09
Missing slash between DOWNLOADS_BASE and DOWNLOADS_URL

masbaehr:
fix wrong url for macos. download did not work because link is like this:
https://downloads.scummvm.orgfrs/scummvm/2.5.1/scummvm-2.5.1-macosx.dmg

### DIFF
--- a/include/Models/DownloadsModel.php
+++ b/include/Models/DownloadsModel.php
@@ -113,7 +113,7 @@ class DownloadsModel extends BasicModel
                 $url = str_replace('{$version}', $version, $url);
             } else {
                 // Construct the URL and fill in the version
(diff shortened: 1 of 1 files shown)


Thunderforge:
It looks like the download was broken as part of #186, specifically the changes to `Constants.php`. The reason this doesn't affect the download links is [this Smarty filter](https://github.com/scummvm/scummvm-web/blob/master/include/Controller.php#L146), which doesn't get run on the recommended download.

Long term, both types of links should be generated consistently, but this PR fixes the immediate problem.

Thanks for doing it! Hope to see more PRs from you!

PR #189, opened 2022-01-10
WEB: Removing Win32 label in extra info

Thunderforge:
Now that we have .exe files for 64-bit Windows, it's no longer accurate to say `Win32 .exe`. This was also unique behavior for Windows; we did not indicate bitness in this way for any other operating system.

## Before
Before

## After
After

### DIFF
--- a/include/Models/DownloadsModel.php
+++ b/include/Models/DownloadsModel.php
@@ -121,10 +121,6 @@ class DownloadsModel extends BasicModel
             $data = ""; //$download->getExtraInfo();
             if (is_array($data)) {
                 $extra_text = $data['size'] . " ";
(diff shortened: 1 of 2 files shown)


include/Models/DownloadsModel.php
```diff
@@ -121,10 +121,6 @@ public function getRecommendedDownload()
             $data = ""; //$download->getExtraInfo();
             if (is_array($data)) {
                 $extra_text = $data['size'] . " ";
-                if ($data['ext'] == '.exe') {
-                    $extra_text = $extra_text . 'Win32 ';
```
Thunderforge, line 125: This code is never actually called due to the commented out L121, but I figure we should future-proof things for when we get around to fixing it.

PR #191, opened 2022-01-10
WEB: Adding file size and extension to recommended downloads

Thunderforge:
More functionality that existed in code, but it didn't work.

Also includes refactoring of the File class.

## Before
Before

## After
After

### DIFF
--- a/include/FileUtils.php
+++ b/include/FileUtils.php
@@ -1,0 +1,118 @@
+<?php
+namespace ScummVM;
+
+/**
+* Utility functions related to files on a file system.
+*/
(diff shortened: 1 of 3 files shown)


Thunderforge:
Codacy isn't happy because I'm "introducing new errors" in new classes. Nevermind that these errors already existed in the old classes, and it's perfectly happy that they are no longer there. I think it's safe to ignore them for now because they already existed.

lotharsm:
Indeed, I see no regressions or newly introduced issues since we already use all the functions codacy complains about.

PR #202, opened 2022-02-06
DATA: Adding links to other ScummVM game mods

Thunderforge:
These mods are made possible by ScummVM.

Links

### DIFF
--- a/data/en/links.yaml
+++ b/data/en/links.yaml
@@ -40,6 +40,25 @@
         these scaler.
       url: >-
         https://web.archive.org/web/20131114143602/http://www.hiend3d.com/hq3x.html
+- name: Game mods made possible by ScummVM
+  notes: >-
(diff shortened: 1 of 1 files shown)


criezy (approved):

eriktorbjorn:
I guess https://wiki.scummvm.org/index.php?title=Loom#Audio_tracks is also a mod, or at least a feature in search of a mod. But maybe it will eventually be covered by the ScummVM Music Enhancement Project.

Thunderforge:
> I guess https://wiki.scummvm.org/index.php?title=Loom#Audio_tracks is also a mod, or at least a feature in search of a mod. But maybe it will eventually be covered by the ScummVM Music Enhancement Project.

I think that’s an interesting idea, but the Links page is more for external websites. If it gets its own page, I don’t see why it shouldn’t be added.

PR #204, opened 2022-02-07
WEB: Upgrading minimum PHP version from 7.3.0 to 8.1.0

Thunderforge:
This would allow us to use Phpfastcache 9.x, which fixes deprecation warnings we get under PHP 8.1.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     }
   ],
   "require": {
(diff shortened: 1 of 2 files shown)


Thunderforge:
Declining because the PHP version is determined by our underlying Debian OS, with the new Debian 11 only having up to 7.4.

PR #205, opened 2022-02-10
IMAGES: Removing fan AGI Screenshots

Thunderforge:
Consensus on Discord was that we don't want future fan games to have their screenshots on the website (with the possible exception of fan games notable enough to have a Wikipedia article). Removing these.

The spreadsheet has already been updated.


Thunderforge:
*Broken Sword 2.5: The Return of the Templars* is the only other obvious fan game we have screenshots for (`company_id == fan`). It doesn't have a Wikipedia page at the moment, but has a few news articles, so I might take at stab at creating one.

We'll just remove the AGI fan games for now and deal with *Broken Sword 2.5* in a future PR.

lotharsm:
I agree that _Broken Sword 2.5_ should be discussed later, since it was even officially supported (e.g. by allowing the use of official character sprites) by _Revolution Software_ and is (at least IMHO) of tremendous quality.

PR #207, opened 2022-02-18
IMAGES: Moving screenshots from company/game to engine/game

Thunderforge:
Includes a recalculation of how screenshot files are counted due to the removal of the `filemask` field.


Thunderforge:
Only four files were changed, which all appear at the top of the "Files changed"

image

The rest were all file moves.

Die4Ever:
this looks good to me, but I'm not familiar with the website code

Thunderforge:
The new Codacy issue is "The use of function glob() is discouraged". However, this is the same way we are doing file matching in other places of the site (such as loading the screenshots). So I think it's better to stay consistent.

include/DataUtils.php
```diff
@@ -128,17 +128,6 @@ private static function convertYamlToOrm()
...
-                        // Add number of shots
-                        $item['num_shots'] = count(
-                            array_filter(
-                                glob("./public_html" . DIR_SCREENSHOTS . "/". $item['filemask']),
-                                function (string $file) {
-                                    return \strpos($file, "_full.");
-                                }
-                            )
-                        );
-                    }
```
Thunderforge, line 141: Previously, we were counting screenshots in this section and saving it to the database as a `num_shots` column for later access. Now we instead count them when it's time to display the screenshots page (and it will be cached so it only happens once per deploy).

PR #208, opened 2022-02-22
WEB: Changing menu link from "Games" to "Freeware Games"

Thunderforge:
"Games" could refer to either 

* Games that are freeware
* Games supported by ScummVM (i.e. Compatibility)

Changing to "Freeware Games" to avoid this ambiguity.

## Before
image

## After
image

### DIFF
--- a/data/en/menus.yaml
+++ b/data/en/menus.yaml
@@ -9,7 +9,7 @@
       href: 'https://forums.scummvm.org/'
     - name: '{#menuMainDownloads#}'
       href: /downloads/
-    - name: '{#menuMainGames#}'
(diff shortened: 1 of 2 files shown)


sev-:
It contains not only freeware games but also addons.

Die4Ever:
Maybe it could be called "Game Downloads"?

Thunderforge:
While "Game Downloads" does avoid the ambiguity with the compatibility table, it doesn't address sev's concern about it being addons too.

Is "Freeware Games & Addons" acceptable? It's longer than most of them, but it's also the most descriptive.

image

PR #212, opened 2022-03-02
WEB: Automatically determine filemask for screenshots

SupSuper:
This is an attempt at fixing the [regression](https://bugs.scummvm.org/ticket/13328) introduced by PR #207 .

The "filemask" column was removed, so now every game screenshot is (incorrectly) duplicated for each variant. This tries to automatically determine the filemask based on the screenshot column data instead. The format is `<game>_<platform>_<language>_<variant>_*`

### DIFF
--- a/include/OrmObjects/Screenshot.php
+++ b/include/OrmObjects/Screenshot.php
@@ -20,7 +20,7 @@ class Screenshot extends BaseScreenshot
     public function getFiles()
     {
         if (!$this->files) {
-            foreach (glob("./" . DIR_SCREENSHOTS . "/" . $this->getGame()->getEngine()->getId() . "/" . $this->getGame()->getId() . "/*") as $file) {
+            foreach (glob("./" . DIR_SCREENSHOTS . "/" . $this->getGame()->getEngine()->getId() . "/" . $this->getGame()->getId() . "/" . $this->getFileMask()) as $file) {
                 if (\strpos($file, "_full.") !== false) {
                     continue;
                 }
@@ -86,6 +86,21 @@ class Screenshot extends BaseScreenshot
         }
 
         return $name;
+    }
+
+    public function getFileMask()
+    {
+        $mask = $this->getGame()->getId() . "_";
+        if ($this->getPlatform()) {
+            $mask .= $this->getPlatform()->getId() . "_";
+        }
+        if ($this->getLanguage()) {
+            $mask .= $this->getLanguage() . "_";
+        }
+        // TODO: Each variant of same platform/language has a sequential ID
+        $mask .= "1_";
+        $mask .= "*";
+        return $mask;
     }
 
     public function __toString()
--- a/include/OrmObjects/Screenshot.php
+++ b/include/OrmObjects/Screenshot.php
@@ -21,12 +21,10 @@ class Screenshot extends BaseScreenshot
     {
         if (!$this->files) {
             foreach (glob("./" . DIR_SCREENSHOTS . "/" . $this->getGame()->getEngine()->getId() . "/" . $this->getGame()->getId() . "/" . $this->getFileMask()) as $file) {
-                if (\strpos($file, "_full.") !== false) {
-                    continue;
-                }
-                // Remove the base folder and extension
+                // Remove the base folder
                 $name = str_replace("./" . DIR_SCREENSHOTS . "/", "", $file);
-                $name = \substr($name, 0, \strlen($name) - 4);
+                // Remove the suffix, eg. "_full.png"
+                $name = \substr($name, 0, \strlen($name) - 9);
                 $this->files[] = [
                     'filename' => $name,
                     'caption' => $this->getCaption(),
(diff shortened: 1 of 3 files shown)


include/OrmObjects/Screenshot.php
```diff
@@ -88,6 +88,21 @@ public function getCaption()
...
+    {
+        $mask = $this->getGame()->getId() . "_";
+        if ($this->getPlatform()) {
+            $mask .= $this->getPlatform()->getId() . "_";
+        }
+        if ($this->getLanguage()) {
+            $mask .= $this->getLanguage() . "_";
+        }
+        // TODO: Each variant of same platform/language has a sequential ID
+        $mask .= "1_";
```
SupSuper, line 101: Still need to determine the variant ID here. Dunno if it would be easier to add a column to the spreadsheet, or calculate it.
(if there are multiple variants for the same platform+language, they are sequentially numbered, eg. `dos_en_1_*`, `dos_en_2_*`, etc)

Thunderforge:
This doesn't seem to have the desired effect. When I load this branch and look at the Loom screenshots, it only shows one for each variant (although two for each PCEngine variant for some reason).

image

Thunderforge:
On closer inspection, it appears that this behavior *is* more or less correct, but there are multiple bugs going on, some of which were not obvious.

First, I saw that there were 18 screenshots for Loom.

image

On the page itself, there are lots of screenshots; current count is 54. It appears that this too is inaccurate because it's just duplicating images.

The correct value should be 9 shots.

image

It appears the 18 shots was counting both the small and the full versions. Thus it should say "9 shots". I created [#13340](https://bugs.scummvm.org/ticket/13340) for addressing that.

So the real issue with this PR is that it results in Loom showing 8 screenshots instead of 9 like it should. Only one DOS English EGA is showing up instead of both.

SupSuper:
@Thunderforge 
> So the real issue with this PR is that it results in Loom showing 8 screenshots instead of 9 like it should. Only one DOS English EGA is showing up instead of both.
Yes this PR is incomplete, I left a TODO for figuring out the variants: https://github.com/scummvm/scummvm-web/pull/212#pullrequestreview-897118658
Feel free to comment if you have any ideas.

Thunderforge:
Although we haven't figured out variants yet, it does fix issues like duplicated files.

I'm okay with merging this now and having the variant fix in a separate PR.

Thunderforge (approved):

include/OrmObjects/Screenshot.php
```diff
@@ -21,12 +21,9 @@ public function getFiles()
...
         if (!$this->files) {
             foreach (glob("./" . DIR_SCREENSHOTS . "/" . $this->getGame()->getEngine()->getId() . "/" . $this->getGame()->getId() . "/" . $this->getFileMask()) as $file) {
-                if (\strpos($file, "_full.") !== false) {
-                    continue;
-                }
-                // Remove the base folder and extension
+                // Remove the base folder and suffix
                 $name = str_replace("./" . DIR_SCREENSHOTS . "/", "", $file);
-                $name = \substr($name, 0, \strlen($name) - 4);
+                $name = \substr($name, 0, \strlen($name) - 9);
```
lephilousophe, line 26: Could you add a comment to explain why 9?

SupSuper: Done

SupSuper:
Ended up just adding a "variant_id" column to the spreadsheet (restoring the old filemasks would require filtering out all the path information), so hopefully this should fix everything for now.

Thunderforge:
This solution appears to be working. For instance, Monkey Island (DOS/English) shows four CD (variant `1`), one Hercules (variant `2`), one CGA (variant `3`), two EGA (variant `4`), and two VGA (variant `5`).

image

lotharsm:
Thank you! Should we merge it as-is?

bluegr:
Looks good to me as well! Nice and clean solution

include/OrmObjects/Screenshot.php
```diff
@@ -20,13 +20,11 @@ class Screenshot extends BaseScreenshot
     public function getFiles()
     {
         if (!$this->files) {
-            foreach (glob("./" . DIR_SCREENSHOTS . "/" . $this->getGame()->getEngine()->getId() . "/" . $this->getGame()->getId() . "/*") as $file) {
-                if (\strpos($file, "_full.") !== false) {
-                    continue;
-                }
-                // Remove the base folder and extension
+            foreach (glob("./" . DIR_SCREENSHOTS . "/" . $this->getGame()->getEngine()->getId() . "/" . $this->getGame()->getId() . "/" . $this->getFileMask()) as $file) {
```
Thunderforge, line 23: Can you rebase? I changed this line in #213

lotharsm: Maybe just change this line here manually, I think it will cause less trouble than le me trying to rebase it here... ;-)

SupSuper: It's not the same line, but I added a check for "_full" in this version as well.

PR #215, opened 2022-03-05
WEB: Escaping special characters in the variant name

Thunderforge:
Fixes an issue where the variant of `2.0F 1987-05-05 5.25"/3.5"` was truncated to `2.0F 1987-05-05 5.25`

## Before
image

## After
image


Thunderforge:
Declining in favor of #216.

PR #218, opened 2022-03-06
WEB: Move screenshot captions to the bottom of the image

SupSuper:
Currently when viewing a screenshot in full, the caption shows at the bottom of the browser. This was probably designed for HD photos, but doesn't look great with our small screenshots:

Before

I moved the caption to show under the screenshot instead:

After


Thunderforge:
I'm not so keen on making modifications to the baguetteBox.js code itself. If we ever applied an update, all your code gets wiped.

According to [this issue](https://github.com/feimosi/baguetteBox.js/issues/61), moving the caption isn't supported. So either we modify the code as you suggested, or we switch to an equivalent library where it is.

Thunderforge:
@SupSuper If you wanted to merge this, you can, with the understanding that any updates may undo your changes.

lotharsm:
Shall we merge this as is?

SupSuper:
Merged with a note about the customization

PR #219, opened 2022-03-10
WEB: Changing grid view to be more responsive

Thunderforge:
Improvements are:

* Long captions are now always visible instead of being cut off
* Differently sized cards no longer cause gaps
* Images are now always at a visible size instead of becoming too small to see

## Before

Before

### Description is not always readable
image

### Gaps can occur with differently sized cards
image

## After
There are no gaps, images stay a consistently readable size, and captions are always shown in full. 

Images are centered on the page, preventing the possibility of a wide gap on the right side

After

### DIFF
--- a/scss/components/_card.scss
+++ b/scss/components/_card.scss
@@ -3,12 +3,10 @@
   flex-direction: column;
   align-items: center;
   justify-content: center;
-  margin-bottom: 20px;
+  margin: 5px;
   padding: 10px;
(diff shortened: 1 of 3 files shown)


lotharsm:
Looks very good to me - thank you very much! As long as there are no objections, we should merge this ASAP 🙂

bluegr:
Looks great, following modern principles. Nice work!

PR #226, opened 2022-03-18
WEB: Adding a caption to random screenshots

Thunderforge:
The caption is a hyperlink that goes to the corresponding game page on https://www.scummvm.org/compatibility/ . This allows users to quickly look up compatibility info as well as see which platforms are supported.

## Before
Home Before
Screenshots Before

## After
Home After
Screenshots After


sev-:
nice improvement

lotharsm:
As sev said, this is a very nice improvement. I think it also helps showcasing all the various games ScummVM supports - the names might be more recognisable than the images themselves (especially on the "more obscure" games).

Merging!

PR #229, opened 2022-04-07
WEB: Sorting same version downloads by priority

Thunderforge:
Uses a new "priority" field to sort downloads that have the same version number instead of an arbitrary order as before.

## Before

Note that the two Raspberry Pi and the two IRIX entries are separated.

Before

## After

The two Raspberry Pi and two IRIX entries are now together.

After

### DIFF
--- a/include/Objects/DownloadsSection.php
+++ b/include/Objects/DownloadsSection.php
@@ -29,14 +29,14 @@ class DownloadsSection extends BasicSection
     {
         if ($item->getCategoryIcon()) {
             $this->items[] = new File($item->toArray(TableMap::TYPE_FIELDNAME), '');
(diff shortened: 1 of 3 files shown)


Thunderforge:
Of course, it's only *after* go through all this work that I realize the reason auto id wasn't working: [this line](https://github.com/scummvm/scummvm-web/blob/master/include/Objects/File.php#L20) has a camelCase `autoId` instead of a snake case `auto_id`.

Switching that to snake cases fixes things. Therefore, I'm going to reject this PR so that the extra complexity of a "Priority" field is not needed.

Thunderforge:
Fixed in 9950eaa063bed397df9e3ce7a4bf928de1fa95b0

PR #230, opened 2022-04-10
WEB: Display file sizes by powers of ten

Thunderforge:
[Per IEC](https://en.wikipedia.org/wiki/Byte#Multiple-byte_units), kilobyte, megabyte, and gigabyte refer to units based on powers of ten (as opposed to powers of two, which are referred to as kibibyte, mebibyte, and gibibyte).

This is used by 

## Before

image

## After

image

### DIFF
--- a/include/FileUtils.php
+++ b/include/FileUtils.php
@@ -31,18 +31,19 @@ class FileUtils
     {
         $path = FileUtils::toAbsolutePathIfOnServer($path);
         // Get the file size, rounded to the nearest kilobyte
-        $file_size = round((@filesize($path) / 1024));
+        // Use powers of ten per IEC standard
+        $file_size = round((@filesize($path) / 1000));
         
-        if ($file_size < 1024) {
-            $file_size = $file_size . "K";
+        if ($file_size < 1000) {
+            $file_size = $file_size . " kB";
         } else {
-            $file_size /= 1024;
+            $file_size /= 1000;
 
-            if ($file_size < 1024) {
-                $file_size = round($file_size, 1) . "M";
+            if ($file_size < 1000) {
+                $file_size = round($file_size, 1) . " MB";
             } else {
-                $file_size /= 1024;
-                $file_size = round($file_size, 2) . "G";
+                $file_size /= 1000;
+                $file_size = round($file_size, 2) . " GB";
             }
         }
         return $file_size;


Thunderforge:
The Codacy issue is flagging existing code (and isn't very helpful because it doesn't suggest an alternative).

sev-:
OH NOES! I always hated this HDD manufacturer lobby with attempts to inflate disk sizes, so now you have no real idea of the true size of your storage!

So, if you absolutely prefer to be 100% precise, then change "KB" to "KiB", but please please do not change to the powers of ten. We're with computers, not in a supermarket.

Thunderforge:
> After looking a second time. Do not even need to change to KiB. "K" is what is used in the modern OSes:

I'm not seeing a "K" in the text that you posted.

Maybe you were saying instead that it was using powers of two? I'm not sure which OS you are using, but being "what is used in the modern OSes" seems at odds with what Wikipedia says:

> Operating systems that use this definition [of powers of ten] include macOS, iOS, Ubuntu, and Debian.

I confirm that the macOS Finder reports it as a power of ten: the Windows installer is reported as 89.1 MB (as in the "After" screenshot") instead of 85 (as in the "Before" screenshot):

image

It also displays KB for smaller files:

image

My main concern is having the website reflect the download experience for end users. It's confusing to have the website report one size, but your operating system report a different one when you download.

criezy:
> My main concern is having the website reflect the download experience for end users. It's confusing to have the website report one size, but your operating system report a different one when you download.

You might have an issue here. Because as far as I know while macOS uses power of tens (as you note), Windows uses power of twos as far as I know, at least in the File Explorer (but incorrectly label them as KB, MB, and so on to make things more confusing).

lotharsm:
> My main concern is having the website reflect the download experience for end users. It's confusing to have the website report one size, but your operating system report a different one when you download.

IIRC it is not even consistent across all modern operating systems - technically, the only accurate way to display the file sizes would be just use bytes without prefix or conversation...

Thunderforge:
So I guess the options are:

* Continue to display as K, M, G (ambiguous for everybody, inaccurate for *nix-based)
* Display as KB, MB, GB (more convenient for everybody, but inaccurate for *somebody*)
* Display as KiB, MiB, GiB (inconvenient for everybody but accurate for all)

It sounds like the path of least resistance may be the final one; it saves the debate over powers of two versus ten and clears up ambiguity.

lotharsm:
What should we do with this one? Continue with the discussion or should we leave it as-is and abandon this proposal for now?

lephilousophe:
I would go the 3rd route.
Figures would be consistent with Windows (largest user base for the website?) while units would be accurate.
It's also how it's displayed on KDE.

As a side note, it seems that Gnome uses power of tens like MacOS and XFCE leaves the choice to the user.

Thunderforge:
Closing in favor of #250, which displays them as @lephilousophe recommends.

PR #231, opened 2022-04-27
IMAGES: Reducing Phantasmagoria 2 screenshots from 30 to 6

Thunderforge:
Per Slack discussion, we don't need so many screenshots. Removing some of the less interesting ones.


Thunderforge:
Here are the six ones currently in use.

image

PR #241, opened 2022-05-16
IMAGES: Adding Blazing Dragons screenshots

SupSuper:
https://bugs.scummvm.org/ticket/12433

Blazing Dragons (Sony Playstation/English (US))


Thunderforge (changes requested):
[From the wiki](https://wiki.scummvm.org/index.php?title=Screenshots):

> For each game, we would like to have a small (say, between 2 and 5) number of screenshots which give users a good idea of what the game looks like to play.

This PR has too many screenshots. Can you cull it down to no more than 10?

sev-:
Not more than 5 per game, please. Wiki is clear here.

Thunderforge (approved):

Thunderforge:
Looks great! I especially like how several of them have English text, clearly indicating that this is the English version.

Thunderforge:
I didn't notice this until later, but the language in the filename was `us`. I had to change it to `en-us` instead ("United States" is not a valid language, but "English - United States" is).

PR #244, opened 2022-05-16
WEB: Adding additional store links

Thunderforge:
Because it's Markdown, we can add whatever we need to in whatever format we need to.

## Before

image

## After

image


Thunderforge:
@lotharsm They are already on the spreadsheet with (almost) everything from [this wiki page](https://wiki.scummvm.org/index.php?title=Where_to_get_the_games) already included.

image

lotharsm:
Very nice! Merging!

PR #245, opened 2022-05-17
WEB: Don't stretch menu banners

SupSuper:
For some reason the menu banners are vertically stretched. I assume it's some obsolete formatting, as there's no point in forcing an image size via CSS, so I removed that rule.

Before:
image

After:
image


lotharsm:
Thank you, merging!

PR #256, opened 2022-06-24
Add DEL to escape characters

einstein95:


sev-:
Thank you!

Thunderforge:
Did we have an issue with this character? Or are we just being preemptive?

sev-:
See the test. It was found on one of the CDs

sev-:
The test here: https://github.com/scummvm/scummvm/pull/4041

And this is a sync with the C++/Python code

PR #260, opened 2022-07-26
Dumper companion: more Mac extensions

mistydemeo:
I was working with a `.cdr` file today, and realized that several Mac-specific file extensions were missing.


sev-:
Thank you!

PR #262, opened 2022-08-03
WEB: Simplifying "Development" category

Thunderforge:
Having two different links going to two separate sections of the GitHub page is a bit redundant.

This is probably a holdover from before pull requests were submitted to GitHub and the source code was in Sourceforge.

## Before

* Pull Requests
* Source Code Tree

image

## After

* Source Code

image


lotharsm:
Instead of using "Source Code", we could simply call the link "GitHub" - I think we can expect people that are interested in contributing that GitHub exists.

Thunderforge:
Done.

image

PR #263, opened 2022-09-08
DATA: Add an old 1.6.0 download for OS X 10.3+

dwatteau:
The current 2.6.0 download for older OS X PowerPC systems requires 10.4 (since there is no easy way to build a C++11 toolchain on older versions), but some retro G3 systems can't run anything newer than OS X 10.3.9, and so some users are looking for a 10.3-compatible release.

User loudmind said on Discord on 2022-08-25 that ScummVM 1.6.0 worked on an old tangerine clamshell iBook from 2000 (this build also has an i386 layer). I can't personally test it, and maybe some later version also works on 10.3, but suggesting ScummVM 1.6.0 for 10.3 users is probably the most helpful thing we can do for now.

### DIFF
--- a/data/en/scummvm_downloads.yaml
+++ b/data/en/scummvm_downloads.yaml
@@ -233,6 +233,16 @@
     enabled: true
     category: scummvm
     subcategory: old
+    category_icon: macosppc
+    user_agent: ''
+    version: 1.6.0
+    url: 'scummvm-{$version}-macosx.dmg'
+    name: 'Mac OS X 10.3+ PPC/Intel 32 bits (without Sparkle) Disk Image'
+    notes: ''
+-
+    enabled: true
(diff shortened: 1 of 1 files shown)


Thunderforge:
This is actually a generated file (see line 1), so any manual edits will get overwritten next time an automated process is run. We have a developer-only data source that goes into making this file.

I've fixed the capitalization issue. I'll confer with other devs to see if we should add a 10.3 listing.

lotharsm:
No objections from my end. I'll close this PR and update the spreadsheet accordingly. I can't test it on my own, so as you said, it is better to provide the user with a version that works for sure - we can also do a version bump later if possible.

Thanks!

lotharsm:
The change is now live on our website.

PR #265, opened 2022-10-01
WEB: Adding engine to compatibility page

Thunderforge:
Adds an "Engine" column to the game compatibility info.

## Before

image

image

## After

image

image

### DIFF
--- a/data/en/strings.json
+++ b/data/en/strings.json
@@ -15,6 +15,7 @@
   "compatibilityDetailsChartTitle": "Game Compatibility Chart",
   "compatibilityDetailsChartColGameFullName": "Game Full Name",
   "compatibilityDetailsChartColGameShortName": "Game Short Name",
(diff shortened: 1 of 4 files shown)


schmurtzm:
I like it a lot !
Like this it is very easier to explain to users how to make the right shortcut, it will help devs to choose which engines to build, it will help end users to diagnose... 
It easy to indicate in the wiki that the right command line is  `scummvm  -p game_directory  Game_Engine:Game_Short_Name`

This way it will avoid problems on command lines with duplicated names like this example on [this issue about Indiana Jones and the Fate of Atlantis](https://forums.scummvm.org/viewtopic.php?p=97463).

So I'm pretty sure that it will be very useful.

sev-:
I actually would suggest to use "<engineid>:<gameid>" because this is how it is shown in the command line.

Thunderforge:
Per Discord feedback, replaced with PR #266.

PR #266, opened 2022-10-02
WEB: Adding "ScummVM ID" to compatibility page

Thunderforge:
This replaces "Game Short Name". When a version newer than 2.2.0 is displayed, the id will be in the format of `engine_id:game_id`. If version 2.2.0 or later is displayed, the id will be in the format of `game_id`.

### Before

Before 260 Table

### After

After 260 Table

### Before

Before 260 Detail

### After

After 260 Detail

### Before

Before 220 Table

### After

After 220 Table

### Before

Before 220 Detail

### After

After 220 Detail


sev-:
Awesome!

PR #267, opened 2022-10-29
WEB: Changing Compatibility ids to engine_id:game_id

Thunderforge:
This change only affects the [compatibility page](https://www.scummvm.org/compatibility/DEV/). Future PRs will address other aspects of the website.

Previously we could have a name collision, as with `qfq1`. Now we can have more unique names: `agi:qfq1` and `sci:qfg1`. This matches the "ScummVM ID" currently visible on the compatibility page.

Note that this PR does change the compatibility URLs:

* **Before:** https://www.scummvm.org/compatibility/DEV/astrochicken/
* **After:** https://www.scummvm.org/compatibility/DEV/sci:astrochicken/

image

### DIFF
--- a/include/DataUtils.php
+++ b/include/DataUtils.php
@@ -23,7 +23,7 @@ class DataUtils
     // filename => sheet id
     const SHEET_IDS = [
         'platforms' => '1061029686',
-        'compatibility' => '854570757',
+        'compatibility' => '1989596967',
         'games' => '1946612063',
         'engines' => '0',
         'companies' => '226191984',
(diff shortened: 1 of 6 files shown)


include/DataUtils.php
```diff
@@ -23,7 +23,7 @@ class DataUtils
     // filename => sheet id
     const SHEET_IDS = [
         'platforms' => '1061029686',
-        'compatibility' => '854570757',
+        'compatibility' => '1989596967',
```
Thunderforge, line 26: You can see this on the data spreadsheet under `Thunderforge WIP - compatibility`. When this change is complete, we will delete the old tab named `compatibility` and rename this tab to the old name.

Thunderforge:
Reverted due to unexpected issues. We'll retry when we get the website working again and can investigate what went wrong.

PR #268, opened 2022-11-12
DUMPER: Add "Always use MacBinary encoding" option.

elasota:
Adds "Always use MacBinary encoding" option, which always dumps files as MacBinary even if they have no resource fork, to keep the HFS type/creator codes intact.

### DIFF
--- a/dumper-companion/src/Dumper.tsx
+++ b/dumper-companion/src/Dumper.tsx
@@ -11,6 +11,7 @@ export type State = {
     lang: Language;
     busy: boolean;
     unicode: boolean;
+    forceMacBinary: boolean;
     logs: ComponentChild[];
     dumpPercent: number;
     lastLogWasDumpPercent: boolean;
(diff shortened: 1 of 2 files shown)


phcoder:
This currently breaks many engines. It needs to be marked with a warning

sev-:
@phcoder could you please elaborate what breaks and in what manner?

Also, what needs to be marked with a warning?

rvanlaar:
Please update the commit message. 

Mention the reasoning why this is needed.

phcoder:
> @phcoder could you please elaborate what breaks and in what manner?

Hades Challenge will need https://github.com/scummvm/scummvm/pull/4372/files or similar.

Kyra engine has similar problem, I had some drafts but wanted to get hadesch one in first. 

Generally the problem is that File.open need to open data fork on game files or be replaced with a function that does
> 
> Also, what needs to be marked with a warning?

PR #272, opened 2022-12-19
WEB: Upgrading minimum PHP version to 7.4.0

Thunderforge:
Also upgrading dependencies, many of which are compatible with 8.x

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
     }
   ],
   "require": {
-    "php": ">=7.3.0",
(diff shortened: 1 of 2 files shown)


lotharsm:
Thank you! Since the website is running on PHP 7.4, I'll merge this and run the update script.

lotharsm:
Looking good. The website works with the updated dependencies and I see no PHP error messages in the logs.

I think we can start with working on PHP 8.x soon:tm:

:-)

PR #274, opened 2022-12-21
WEB: Upgrade PHP from 7.4 to 8.2

Thunderforge:
Everything seems to be working properly thanks to the updated dependencies.

Coordinating with @lotharsm to merge this at the same time that we switch over to a VM with PHP 8.2.


lotharsm:
Thank you!

In theory, I can switch over the PHP versions pretty easily on the same VM. However, what's left to do is migrating the Website and planet.scummvm.org to two seperate PHP-FPM pools, since planet.scummvm.org is not compatible with PHP 8.x and I have little hope that it is still maintened [1].

[1] https://github.com/moonmoon/moonmoon

lotharsm:
Thank you, merging!

PR #287, opened 2023-04-19
WEB: Redirecting to compatibility page when no game id exists

Thunderforge:
Part of the migration to using full game ids (i.e. `monkey` becomes `scumm:monkey`).

## Before

https://www.scummvm.org/compatibility/DEV/foobar/ results in an error page

## After

https://www.scummvm.org/compatibility/DEV/foobar/ redirects to https://www.scummvm.org/compatibility


Thunderforge:
Codacy doesn't like the way we're forcing a redirect, but as best I can tell, that is [the accepted way to do it](https://stackoverflow.com/a/768472). I think we can ignore Codacy.

PR #300, opened 2023-06-04
WEB: fix Vampire Diaries icon

lephilousophe:
Vampire Diaries icon is misplaced


Thunderforge (commented):
I confirm that before this PR, `vampirediares.png` does not show up on the screenshots page, and after this PR it does.

But I'm not understanding what the purpose of the adding a generic linux icon. What does this fix?

lephilousophe:
Generic Linux icon is not a fix.
It would be for a future (yet to come) package output by Buildbot which works on every Linux.

The website and Buildbot share the same icons and Buildbot (ab)uses the CSS and PNG hosted on the website.

Anyway, now you make me think about it again, as the package will be an AppImage, maybe I could insert an AppImage logo instead.
I will do it later.

Thunderforge:
I think it would be better for this PR to be just the Vampire Diaries fix and another PR to be your improvements to Linux.

lephilousophe:
Please note that the PR was made of 2 atomic commits. One with the icon fix and one with the new icon.

As I will change this icon anyway, I only kept the fix icon commit.

Thunderforge (approved):
Thanks!

PR #322, opened 2023-08-13
WEB: Removing GOG affiliate link

Thunderforge:
Per Discord discussion, the legacy affiliate program is ending.

## Before

Before

## After

After


lotharsm:
Thank you, merging!

PR #329, opened 2023-08-25
WEB: Adding ZOOM Platform link to Compatibility pages

Thunderforge:
## Before
Before

## After
After


lotharsm:
Thank you, merging!

PR #337, opened 2023-09-24
IMAGES: Reduced file size of "Screenshot game icon" pngs

GandalfTheWhite80:
Used pngcrush/optipng/advpng/advdef to reduce the file size of these pngs.  Notes:
- many icons have now a lower bit depth. This should not be a problem
- Some chunks maybe

Files became 1/3 smaller on average


GandalfTheWhite80:
As per discussion these icons are  combined into a spritesheet. No need for optimization.

PR #387, opened 2024-04-13
WEB: Add Thimbleweed Park announcement

scemino:


data/en/news/20240413.markdown
```diff
@@ -0,0 +1,16 @@
+---
+title: A walk in the park with ScummVM and Thimbleweeds
+author: scemino
+date: 1712994918
+---
+
+1987 in the small town of Thimbleweed Park, two officers-a-reno are investigating a murder.
```
einstein95, line 7: Grammatical error here, you may want "In 1987, in the small town of Thimbleweed Park," or even "1987: In the small town of Thimbleweed Park,"

scemino: Is it OK?

einstein95: Didn't see your reply but yup it is

PR #411, opened 2024-09-06
WEB: Improve OpenGraph and SEO headers and fix news title

lephilousophe:
Improve the OpenGraph result by diversifying the description and the title.
This is done by adding more template parameters that each Controller sets to the proper value depending on the context.
This allows a better rendering than in #393.


lwcorp:
"better" is a subjective term. For example, I've fixed `<title>` while you didn't. I've also found a way to dynamically generate a meta description on the fly, which could be served as `else` to your check of `if isset($description)`.
Other differences are in [here](https://github.com/scummvm/scummvm-web/pull/393#issuecomment-2334720986) - as suggested there, why not just editing it to add your features there, and thus enjoy both my features and yours?

lephilousophe:
This is updated to add subtitle to the page title.
In addition, the news page title gains a dedicated string. There is now Home for the home page and Latest developments (ripped from the content title) for the news archive page.
This title can be changed afterwards.

Thunderforge (approved):

lephilousophe:
Merging now.
I just added a change so that the page description when showing a game compatibility is the compatibility status.
This will allow to link a game compatibility page and get a nice "Game has no known issue" in the preview.

lwcorp:
Though it's merged it seems it's still not live, do you know the ETA?
BTW, out of curiosity how were you able to get write access, which is required for assigning reviewers and merging?

PR #423, opened 2024-10-05
JANITORIAL: Fix web page typos

raziel-:


lephilousophe:
Thanks.
I just pushed a squashed version of your PR.
You fixed some autogenerated files so I had to fix it manually.
I also skipped the fix on the press article on Gameboy as it was the original title.

raziel-:
ok, sorry for the extra work

thank you

PR #479, opened 2025-05-11
deps: downgrade zip.js to 2.7.56

mistydemeo:
2.7.57 and newer have a bug which creates incorrect subdirectory permissions when ZIP files are opened on Unix systems (macOS, Linux, etc.). The version pin can be removed once there's a new version with fixed permissions.

If you use a Unix OS, you can test this on the currently-live version of dumper companion by trying to extract an ISO containing subdirectories. The subdirectories will be created as `-rw-r--r--@`, with no executable permissions, and so both the terminal and GUI file browsers like Finder will refuse to let the user access those directories.

See: https://github.com/gildas-lormeau/zip.js/issues/567

### DIFF
--- a/dumper-companion/package-lock.json
+++ b/dumper-companion/package-lock.json
@@ -8,7 +8,7 @@
       "name": "dumper-companion",
       "version": "0.0.1",
       "dependencies": {
-        "@zip.js/zip.js": "^2.7.60",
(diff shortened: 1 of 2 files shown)


mistydemeo:
Closing in favour of #480.

lephilousophe:
I applied my suggested workaround to avoid downgrade.
Thank you.

PR #480, opened 2025-05-13
deps: upgrade zip.js

mistydemeo:
This fixes the directory permission bug that broke ZIP files on Mac/Linux. Dumper Companion needs to use either 2.7.61+ or 2.7.56 and lower.

Replacement for #479 now that a fixed release is out.

### DIFF
--- a/dumper-companion/package-lock.json
+++ b/dumper-companion/package-lock.json
@@ -8,7 +8,7 @@
       "name": "dumper-companion",
       "version": "0.0.1",
       "dependencies": {
-        "@zip.js/zip.js": "^2.7.60",
+        "@zip.js/zip.js": "^2.7.61",
         "preact": "^10.26.5",
         "punycode": "^2.3.1",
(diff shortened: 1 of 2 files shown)


lephilousophe:
I took some time to investigate the problem yesterday and the new zip.js version is still not correct as it now misses the directory bit in the permission mask.
In addition, the mask misses the DOS attributes which makes it incompatible with older Zip decompressors.
I have another fix in mind: forcing the creation of msdos compatible archives.
I also plan to propose a proper patch to zip.js.
This will have to wait the end of week.

mistydemeo:
In the meantime, I'll reopen the other PR so we can at least revert to the old version and have dumper companion working in the interim.

lephilousophe:
I issued the PR https://github.com/gildas-lormeau/zip.js/pull/569 to fix the bug.
So I will wait before upgrading.

lephilousophe:
The update have been picked by dependabot so we can close this PR.
Thanks!

PR #523, opened 2025-08-12
WEB: Improve static analysis

lephilousophe:
I did several passes of PHPStan on our code and managed to increase its level of scanning.
This has the benefit of having code theoretically more resilient but at the same time, as I had to add typing to PHP function arguments, this also makes PHP checks the types dynamically and can trigger more exceptions.
I think I tested all endpoints and it seems to work well but that's not 100% sure.

I also added a Github CI check to have a first glimpse on PR quality. I expect this check to allow to be more confident about Dependabot PRs (in addition to test locally).
This allowed me to spot that the Glue sprite generator we use is completely broken on Debian/Ubuntu because Python Pillow library got updated but not Glue. I patch it in the CI but that's far from ideal.

I tried to find some equivalent but failed to find something maintained: it seems that CSS sprites are not a thing anymore?


lephilousophe:
I merged it manually

